PR #16, opened 2016-11-07
update package

gaocegege:
Sorry, I have imported a bug about version

update the packages now.

PTAL


superxi911:
LGTM

PR #44, opened 2016-11-10
delete the repeated code

xiaomuqiao:
delete the repeated code to avoid sending email twice


xiaomuqiao:
PTAL

gaocegege:
LGTM, I would merge it now.

PR #52, opened 2016-11-15
add test cases

### LINKED ISSUE #43: add more unit test cases
Now code coverage of cyclone is a little low 😕 , so we need more test cases.

gaocegege:
fixes #43 

### DIFF
--- a/docker/manager.go
+++ b/docker/manager.go
@@ -31,10 +31,10 @@ import (
 
 // Manager manages all docker operations, like build, push, etc.
 type Manager struct {
-	client     *docker_client.Client
-	registry   string
-	authConfig *AuthConfig
-	endPoint   string
+	Client     *docker_client.Client
+	Registry   string
+	AuthConfig *AuthConfig
+	EndPoint   string
 }
 
 // NewManager creates a new docker manager.
@@ -52,10 +52,10 @@ func NewManager(endpoint string, certPath string, registry api.RegistryCompose) 
 		}
 
 		return &Manager{
-			client:     client,
-			registry:   registry.RegistryLocation,
-			authConfig: authConfig,
-			endPoint:   endpoint,
+			Client:     client,
+			Registry:   registry.RegistryLocation,
+			AuthConfig: authConfig,
+			EndPoint:   endpoint,
 		}, nil
 	}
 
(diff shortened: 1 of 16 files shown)


gaocegege:
I have tried to run e2e tests in travis but fails because of the time limit.

gaocegege:
PTAL

gaocegege:
PTAL @superxi911

api/rest/worker_node.go
```diff
@@ -23,12 +23,14 @@ import (
...
 	"github.com/caicloud/cyclone/pkg/log"
+	"github.com/caicloud/cyclone/pkg/osutil"
 	"github.com/caicloud/cyclone/store"
 	"github.com/emicklei/go-restful"
 	docker_client "github.com/fsouza/go-dockerclient"
 )
 
-const WORKER_NODE_DOCKER_VERSION = "1.10.1"
+// WORKER_NODE_DOCKER_VERSION defines the required docker version in worker node.
+const WORKER_NODE_DOCKER_VERSION = "WORKER_NODE_DOCKER_VERSION"
```
superxi911: I think we should restrict it by code, not by ENV

docker/manager.go
```diff
@@ -31,10 +31,10 @@ import (
...
 // Manager manages all docker operations, like build, push, etc.
 type Manager struct {
-	client     *docker_client.Client
-	registry   string
-	authConfig *AuthConfig
-	endPoint   string
+	Client     *docker_client.Client
+	Registry   string
+	AuthConfig *AuthConfig
+	EndPoint   string
```
superxi911: Why expose these vars? We have functions to get them, such as: GetDockerAuthConfig, GetDockerRegistry. If you expose them, them can be modified

api/rest/worker_node.go
gaocegege: For example, if we wanna change the supported docker version in worker node, we have to re-compile the code.

It's not cool.

docker/manager.go
gaocegege: just for test, we can not mock this manager if we don't expose them.

api/rest/worker_node.go
gaocegege: I changed here because travis's docker version is 1.12.0, and this version could pass e2e tests.

gaocegege: And now the default is 1.10.1, I think it would be OK.

We could add some tips in doc, for example "This is just for test."

docker/manager.go
xiaomuqiao: Would exposed the vars violate the golang's coding style?  if yes,  i think can delete the functions, as GetDocker***

api/rest/worker_node.go
gaocegege: OK

gaocegege:
PTAL

superxi911:
LGTM

Issue #53, opened 2016-11-15
./scripts/build-image.sh failed


wanghaoran1988:
./scripts/build-image.sh 
+++++ Start building cyclone server
can't load package: package github.com/caicloud/cyclone: open /go/src/github.com/caicloud/cyclone: permission denied

wanghaoran1988:
docker run -u 0 -t -i -v `pwd`:/go/src/github.com/caicloud/cyclone \
-e GOPATH=/go:/go/src/github.com/caicloud/cyclone/vendor golang:1.6-alpine sh
/go # cd /go/src/github.com/caicloud/cyclone/vendor
/go/src/github.com/caicloud/cyclone/vendor # ls -rlt
ls: can't open '.': Permission denied
total 0

wanghaoran1988:
We need to disable selinux if we run on centos/redhat using sudo setenforce 0

gaocegege:
Thanks for the issue and PR, we would test it ASAP :)

PR #54, opened 2016-11-15
 disable the selinux before build the image

wanghaoran1988:
fix  https://github.com/caicloud/cyclone/issues/53

### DIFF
--- a/scripts/build-image.sh
+++ b/scripts/build-image.sh
@@ -10,6 +10,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+source "$(dirname "${BASH_SOURCE}")/lib/common.sh"
+
 ROOT=$(dirname "${BASH_SOURCE}")/..
 
 function usage {
@@ -24,7 +26,6 @@ function usage {
   echo -e "Environment variable:"
   echo -e " PUSH_TO_REGISTRY     \tPush images to caicloud registry or not, options: Y or N. Default value: ${PUSH_TO_REGISTRY}"
 }
-
 # -----------------------------------------------------------------------------
 # Parameters for building docker image, see usage.
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


scripts/build-image.sh
```diff
@@ -24,7 +24,13 @@ function usage {
   echo -e "Environment variable:"
   echo -e " PUSH_TO_REGISTRY     \tPush images to caicloud registry or not, options: Y or N. Default value: ${PUSH_TO_REGISTRY}"
 }
-
+function disable-selinux() {
+  if selinuxenabled && [ "$(getenforce)" = "Enforcing" ]; then
```
gaocegege: I run your PR in OS X 10.12.1, and it raises an error:

```
./scripts/build-image.sh: line 28: selinuxenabled: command not found
```

maybe we should define it before this function :)

scripts/build-image.sh
```diff
@@ -65,10 +71,11 @@ echo "+++++ Start building cyclone server"
 cd ${ROOT}
 
 # Build cyclone comtainer.
+disable-selinux
```
gaocegege: By the way, could you add some comments on it :)

scripts/build-image.sh
wanghaoran1988: @gaocegege Sorry, don't have an mac on hand, I am using fedora, will have an update, when a mac in hand.

gaocegege: No worry :) 
Is `selinuxenabled ` a built-in command in fedora?

wanghaoran1988: selinuxenabled - tool to be used within shell scripts to determine if selinux is enabled

superxi911: +1
> OK, I think, we should bypass this function call when the script is running in OS X and other OSes which don't have SELinux features :)

wanghaoran1988: ok

gaocegege:
LGTM, it works for me. PTAL @superxi911 
😄

superxi911:
LGTM

PR #63, opened 2016-11-19
make the error message more human readable

gaocegege:
PTAL @superxi911 


xiaomuqiao:
LGTM

PR #67, opened 2016-11-21
add log to log file and adjust the order of the  vulnerabilities from  clair checking

xiaomuqiao:
1、 add log in log file when the specified  caicloud.yaml is not exsited 
2、 adjust the order of the  vulnerabilities from  clair checking


superxi911:
LGTM

PR #68, opened 2016-11-23
fix unit test cases

gaocegege:
PTAL @superxi911 @xiaomuqiao 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 
-# sudo: required
+sudo: required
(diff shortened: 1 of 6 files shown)


superxi911:
`mkdir: cannot create directory `/logs': Permission denied`

gaocegege:
I have fixed it now. Unit test could be passed in travis and cyclone itself now.
PTAL @superxi911

superxi911:
LGTM

PR #72, opened 2016-11-24
add  parsing yaml log in caicloud.yaml ; increase more check when clone code

xiaomuqiao:
1、increase the check to prevent crashes
2、set service status according event's status
3、add log info about parsing yaml

### DIFF
--- a/worker/vcs/manager.go
+++ b/worker/vcs/manager.go
@@ -60,18 +60,18 @@ func getUrlwithToken(url, subvcs, token string) string {
 	switch subvcs {
 	case api.GITHUB:
 		if token == "" {
-			log.Warn("subvcs is github, but the token is empty, there meybe exist error")
+			log.Warn("Sub VCS is GitHub, but the token is empty, maybe there is an error")
 			break
 		}
 		position := strings.Index(url, "github.com")
 		if position == -1 {
-			log.Warn("subvcs is github and the token is not empty, but the url dosen't contain github.com")
+			log.Warn("subvcs is github and the token is not empty, but the url dosen't contain 'github.com'")
 			break
 		}
 		url = insert(url, token+"@", position)
 	case api.GITLAB:
 		if token == "" {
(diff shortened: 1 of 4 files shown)


worker/vcs/manager.go
```diff
@@ -59,9 +59,21 @@ func insert(url, insertion string, index int) string {
 func getUrlwithToken(url, subvcs, token string) string {
 	switch subvcs {
 	case api.GITHUB:
+		if token == "" {
+			log.Warn("subvcs is github, but the token is empty, there meybe exist error")
```
gaocegege: I think the log could be "Sub VCS is GitHub, but the token is empty, maybe there is an error."

worker/vcs/manager.go
```diff
@@ -59,9 +59,21 @@ func insert(url, insertion string, index int) string {
...
+			break
+		}
 		position := strings.Index(url, "github.com")
+		if position == -1 {
+			log.Warn("subvcs is github and the token is not empty, but the url dosen't contain github.com")
```
gaocegege: I think the log could be "Sub VCS is GitHub, but the URL doesn't contain 'github.com'"

worker/vcs/manager.go
```diff
@@ -59,9 +59,21 @@ func insert(url, insertion string, index int) string {
...
+			break
+		}
 		url = insert(url, token+"@", position)
 	case api.GITLAB:
+		if token == "" {
+			log.Warn("subvcs is gitlab, but the token is empty, there meybe exist error")
```
gaocegege: as above.

worker/vcs/manager.go
xiaomuqiao: ok

worker/vcs/manager.go
xiaomuqiao: done

worker/vcs/manager.go
xiaomuqiao: done

superxi911:
LGTM

PR #74, opened 2016-11-24
remove docker network

gaocegege:

### DIFF
--- a/docker/manager.go
+++ b/docker/manager.go
@@ -245,11 +245,6 @@ func (dm *Manager) GetAuthOpts() (authOpts docker_client.AuthConfigurations) {
 	return authOpts
 }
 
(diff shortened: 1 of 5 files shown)


gaocegege:
PTAL @superxi911 @xiaomuqiao

gaocegege:
It seems that cyclone meets an error.

superxi911:
We maybe need to lengthen the [time](https://github.com/caicloud/cyclone/blob/master/websocket/websocket_test.go#L42) for waiting websocket server start up.  How about 5ms？

gaocegege:
OK but I'm wondering why cyclone is failed.

gaocegege:
PTAL

superxi911:
LGTM

PR #75, opened 2016-11-25
Add service name to description of GitHub PR status.

superxi911:


superxi911:
One repo may have multi services. For example, Cyclone may have unit-test and e2e-test. Them will appear here at the same. We need to distinguish them.

superxi911:
@gaocegege @xiaomuqiao PTAL

xiaomuqiao:
LGTM

PR #77, opened 2016-11-27
fix error for the defaut values of some environment variables

supereagle:


gaocegege:
Hi, Thanks for your contribution first, But I think maybe you should set these env vars in [Dockerfile](https://github.com/caicloud/cyclone/blob/master/Dockerfile).

:)

superxi911:
I agree with @gaocegege . It's unnecessary. Thank you😊

supereagle:
I found this error when run the tests in package `notify/provider`. Run tests after skipped the test case `TestReadConfig`, there will be panic error. The root cause is that these env variables will not be set when skip `TestReadConfig`, then it will use their default values, which are not correct.

At the same time, I think these test cases should be independent and isolated, have no effects on each other.

xiaomuqiao:
yes, your proposal is great. so could you modify these test cases  in package notify/provider?

supereagle:
@xiaomuqiao  Yes. I will commit these changes in another PR.

gaocegege:
@supereagle Cool! 👏

supereagle:
If isolate these test cases, this PR is needed, as these default values are needed in these test cases.  @gaocegege  Could you merge this PR?

gaocegege:
you could

gaocegege:
@xiaomuqiao @supereagle @superxi911 WDYT?

superxi911:
I agree with @gaocegege .

supereagle:
I also agree.  According to the [code](https://github.com/caicloud/cyclone/blob/master/Dockerfile#L15) in Dockerfile, these default values are `CORRECT` values in the image. I think we can closed this PR.

gaocegege:
Yeah thank you for your advice. would you help us to fix these test cases? We'd appreciate it :)

supereagle:
To isolate these test cases, I will unset these envs at the end of `TestReadConfig`, and set wrong envs at the head of `TestSendEmailWithWrongConfig`. 

@gaocegege  @superxi911  @xiaomuqiao   WDYT?

gaocegege:
Great, test cases are all written on my own, and I was rush into increasing the coverage at that time. If you find more bugs in unit test or end-to-end test cases, feel free to fix them :)

PR #78, opened 2016-11-28
add latest tag for image name which dosen't have tag name

xiaomuqiao:
fix issue about pulling all tags when  image name which dosen't have tag name

### DIFF
--- a/docker/manager.go
+++ b/docker/manager.go
@@ -25,6 +25,7 @@ import (
 	"github.com/caicloud/cyclone/pkg/filebuffer"
 	"github.com/caicloud/cyclone/pkg/log"
 	steplog "github.com/caicloud/cyclone/worker/log"
+	"github.com/docker/distribution/reference"
 	"github.com/docker/docker/builder/dockerfile/command"
 	docker_parse "github.com/docker/docker/builder/dockerfile/parser"
 	docker_client "github.com/fsouza/go-dockerclient"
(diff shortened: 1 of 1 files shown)


xiaomuqiao:
PTAL

docker/manager.go
```diff
@@ -85,6 +86,16 @@ func NewManager(endpoint string, certPath string, registry api.RegistryCompose)
 
 // PullImage pulls an image by its name.
 func (dm *Manager) PullImage(imageName string) error {
+	repo, err := reference.Parse(imageName)
+	if err != nil {
+		log.Errorf("imagename parse error: %v", err)
```
superxi911: Why not return?

docker/manager.go
```diff
@@ -85,6 +86,16 @@ func NewManager(endpoint string, certPath string, registry api.RegistryCompose)
...
+	} else {
+		tagged, ok := repo.(reference.Tagged)
+		if !ok || tagged.Tag() == "" {
```
superxi911: I think it need to return err when "ok==false"

docker/manager.go
xiaomuqiao: I deliberately do not quit，i think it is not so serious

xiaomuqiao: But it can also return here

xiaomuqiao:
this pr is from "from unknown repository", cloned it. i will submit again.

PR #80, opened 2016-11-29
add latest tag for image name which dosen't have tag name

xiaomuqiao:
fix issue about pulling all tags when image name which dosen't have tag name

### DIFF
--- a/docker/manager.go
+++ b/docker/manager.go
@@ -25,6 +25,7 @@ import (
 	"github.com/caicloud/cyclone/pkg/filebuffer"
 	"github.com/caicloud/cyclone/pkg/log"
 	steplog "github.com/caicloud/cyclone/worker/log"
+	"github.com/docker/distribution/reference"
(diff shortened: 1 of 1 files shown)


docker/manager.go
```diff
@@ -85,6 +86,17 @@ func NewManager(endpoint string, certPath string, registry api.RegistryCompose)
 
 // PullImage pulls an image by its name.
 func (dm *Manager) PullImage(imageName string) error {
+	repo, err := reference.Parse(imageName)
+	if err != nil {
+		log.Errorf("imagename parse error: %v", err)
+		return err
+	} else {
```
superxi911: Needn't else

gaocegege:
There is a new dependency, I'm not sure if we should update vendor 🤔

xiaomuqiao:
the new "reference" dependency have been existed in vendor, @gaocegege

superxi911:
LGTM

PR #82, opened 2016-11-29
support cancel building job

xiaomuqiao:
make changes to support cancel building job manually
Also fix issue https://github.com/caicloud/cyclone/issues/83

### DIFF
--- a/event/manager.go
+++ b/event/manager.go
@@ -27,11 +27,12 @@ import (
 	"github.com/caicloud/cyclone/pkg/log"
 	"github.com/caicloud/cyclone/remote"
 	"github.com/caicloud/cyclone/resource"
+	"github.com/caicloud/cyclone/store"
 	"golang.org/x/net/context"
 )
 
 const (
-	// unfinished event dir path in etcd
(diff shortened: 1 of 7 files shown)


event/manager.go
```diff
@@ -158,6 +158,17 @@ func eventChangeHandler(event *api.Event, preEvent *api.Event) {
 			log.Errorf("delete finished event err: %v", err)
 		}
 	}
+
+	// make the state of version as running, the frontend will be based on this state to decide whether to show cancel button.
+	if preEvent.Status == api.EventStatusPending && event.Status == api.EventStatusRunning && event.Operation == CreateVersionOps {
```
superxi911: I think the event should can be canceled after it accepted.

xiaomuqiao: but now  the front end is based on the version status to determine whether to display the cancel button

superxi911: You can set the status of version to  "api.VersionRunning" when the event create.

xiaomuqiao: ok

superxi911:
LGTM

PR #90, opened 2016-12-02
add badge for service

gaocegege:

### DIFF
--- a/api/rest/service.go
+++ b/api/rest/service.go
@@ -27,6 +27,14 @@ import (
 	"github.com/emicklei/go-restful"
 )
 
+const (
+	// successBadge is the badge showed when version creation is successfully.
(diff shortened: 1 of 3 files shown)


gaocegege:
PTAL @superxi911 @xiaomuqiao 
This feature needs to re-deploy cyclone. And the badge would return the latest version status.

xiaomuqiao:
LGTM

xiaomuqiao:
The next is that testing with cyclone web.

superxi911:
LGTM

api/rest/service.go
```diff
@@ -27,6 +27,14 @@ import (
...
+
+	// failureBadge is the badge showed when version creation is failed.
+	failureBadge = `<svg xmlns="http://www.w3.org/2000/svg" width="80" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="80" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h37v20H0z"/><path fill="#e05d44" d="M37 0h43v20H37z"/><path fill="url(#b)" d="M0 0h80v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="18.5" y="15" fill="#010101" fill-opacity=".3">build</text><text x="18.5" y="14">build</text><text x="57.5" y="15" fill="#010101" fill-opacity=".3">failing</text><text x="57.5" y="14">failing</text></g></svg>`
```
xiaomuqiao: i think  cyclone only needs to provide results, how display the result  can be  implemented by a third party. Just a   little idea.

gaocegege: We don't have third party support now and we should have a solution now, so I think we could keep it now.

Issue #73, opened 2016-11-24
Send heart beat  fail after cyclone server is restarted


xiaomuqiao:
Describe:
      When i restarted the cyclone server, the  worker can't send heart beat and log info to server through  websocket.

superxi911:
Cyclone worker need to reconnect to server when the websocket disconnect. I'll fix it.

PR #109, opened 2016-12-09
fix: redail when worker disconnect from log server

superxi911:
fix [#73](https://github.com/caicloud/cyclone/issues/73)


superxi911:
@xiaomuqiao @zoumo PTAL

zoumo:
LGTM

PR #171, opened 2017-09-04
add list metadata and method, move struct error response to tpyes.go

### LINKED ISSUE #169: Add list response metadata and method
Successful collection GET operation MUST return metadata with 'total' and 'items' array, i.e.

{
    "metadata": {
        "total": 100,
        "itemsLength": 10
    },
    "items": [ {}, {}, {} ]
}

Collection GET (A.K.A. List APIs)

SHOULD support pagination query: GET /api/v0/items?start=&queryName2=

- start: (number, optional) - query start index, default 0.
- limit: (number, optional) - specify the number of records, default +Inf, return all.

zz-bmj:
## Fixes   #169 

**Changes**
   
  - [x] add list metadata and method
  - [x] move struct error response to pkg/api/tpyes.go
  - [x] support pagination query

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -269,3 +269,36 @@ type GeneralStageStatus struct {
 	StartTime time.Time `bson:"startTime,omitempty" json:"startTime,omitempty" description:"start time of the stage"`
 	EndTime   time.Time `bson:"endTime,omitempty" json:"endTime,omitempty" description:"end time of the stage"`
 }
+
+// ListMeta represents metadata that list resources must have.
+type ListMeta struct {
+	Total       int `json:"total" description:"total items count"`
+	ItemsLength int `json:"itemsLength" description:"returned items count"`
+}
+
+// ListResponse represents a collection of some resources.
+type ListResponse struct {
+	Meta  ListMeta    `json:"metadata" description:"pagination object"`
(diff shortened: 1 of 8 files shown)


pkg/api/types.go
```diff
@@ -269,3 +269,22 @@ type GeneralStageStatus struct {
 	StartTime time.Time `bson:"startTime,omitempty" json:"startTime,omitempty" description:"start time of the stage"`
 	EndTime   time.Time `bson:"endTime,omitempty" json:"endTime,omitempty" description:"end time of the stage"`
 }
+
+// ListMeta represents metadata that list resources must have.
+type ListMeta struct {
```
supereagle: `ListMeta` is not needed as it may not be used independently, directly put its fields into `ListResponse`.

pkg/api/types.go
```diff
@@ -269,3 +269,22 @@ type GeneralStageStatus struct {
...
+	Total       int `json:"total,omitempty" description:"total items count"`
+	ItemsLength int `json:"itemsLength,omitempty" description:"returned items count"`
```
supereagle: Both `Total` and `ItemsLength` is required, can not be `omitemtpy`. `Items` can be `omitemtpy`.

PR #177, opened 2017-09-07
fix the bug with wrong collection name

zz-bmj:
so sorry, It's my mistake that hasn't checked the code enough.


supereagle:
LGTM

PR #186, opened 2017-09-11
fix Image name to match new requirement

### LINKED ISSUE #185: Image name need to be changed
now, the image name needs to be like "docker-address/userName/pipelineName: timestamp-commit_id" as a default image name if the user does not input image name.

zz-bmj:
## Fixes #185

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -187,27 +187,12 @@ func (worker *Worker) createVersion(vcsManager *vcs.Manager, event *api.Event) {
 	destDir := vcsManager.GetCloneDir(&event.Service, &event.Version)
 	event.Data["context-dir"] = destDir
 
-	var imageName, tagName string
-
-	if event.Service.ImageName == "" {
-		imageName = fmt.Sprintf("%s/%s/%s", dockerManager.Registry,
-			strings.ToLower(event.Service.Username), strings.ToLower(event.Service.Name))
-		tagName = fmt.Sprintf("%s-%d", event.Version.Commit, time.Now().Unix())
-	} else {
-		if arr := strings.Split(event.Service.ImageName, ":"); len(arr) == 2 {
-			imageName, tagName = arr[0], arr[1]
-		} else {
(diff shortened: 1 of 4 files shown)


worker/worker.go
```diff
@@ -186,9 +186,23 @@ func (worker *Worker) createVersion(vcsManager *vcs.Manager, event *api.Event) {
...
-	event.Data["image-name"] = fmt.Sprintf("%s/%s/%s", dockerManager.Registry,
-		strings.ToLower(event.Service.Username), strings.ToLower(event.Service.Name))
-	event.Data["tag-name"] = event.Version.Name
+
+	var imageName, tagName string
+
+	if "" == event.Service.ImageName {
+		imageName = fmt.Sprintf("%s/%s/%s", dockerManager.Registry,
+			strings.ToLower(event.Service.Username), strings.ToLower(event.Service.Name))
+		tagName = fmt.Sprintf("%d-%s", time.Now().Unix(), event.Version.Commit)
```
supereagle: What about if `event.Version.Commit` is empty? 
The format should be `{short commit}-{timestamp}`, like `7654321-170401040120`.

worker/worker.go
```diff
@@ -186,9 +186,23 @@ func (worker *Worker) createVersion(vcsManager *vcs.Manager, event *api.Event) {
...
+	} else {
+		if arr := strings.Split(event.Service.ImageName, ":"); len(arr) == 2 {
+			imageName, tagName = arr[0], arr[1]
+		} else {
+			setEventFailStatus(event, err.Error())
```
supereagle: Here the `err` is nil, must new an error to illustrate the fail reason.

worker/worker.go
zz-bmj: done

worker/worker.go
zz-bmj: done

zz-bmj:
@supereagle PTAL

supereagle:
LGTM

PR #191, opened 2017-09-12
Fix cascaded deletion error for pipelines and projects

### LINKED ISSUE #184: Fail to delete the pipeline which has no pipeline records
**Environment**
- branch: v1-preview

**What happened**
Error when delete the pipeline which has no pipeline records:
`[2017-09-10 22:20:26] [ERROR] [pipeline.go:141] Fail to delete all pipeline records of the pipeline pipeline1 in the project project1 as not found`

**What you expected to happen**
- Correctly delete the pipeline which has no pipeline records.

supereagle:
## Fixes #184 

**Changes**  
  - [x] Not return error for deletion when cascaded resources do not exist


zz-bmj:
LGTM

PR #240, opened 2017-11-07
feat(server): set default name when name empty in creating pipeline record

LiangXugang:
**Changes**
   
   - [ ] feature

set '$createTime|$commitID' instead of random string as default pipelineRecord.name 

**Reviewers**

  - @supereagle please review

    
**Checklist**
   
   - [x] Rebased/mergable
   - [ ] Tests pass
   - [ ] Issue/task done

### DIFF
--- a/api/conversion/conversion.go
+++ b/api/conversion/conversion.go
@@ -208,12 +208,6 @@ func ConvertPipelineParamsToVersion(performParams *newapi.PipelinePerformParams)
 	stagesStr := strings.Join(performParams.Stages, ",")
 	version.Operation = api.VersionOperation(strings.Replace(stagesStr, "imageRelease", "publish", 1))
 
-	if performParams.Name != "" {
-		version.Name = performParams.Name
-	} else {
-		version.Name = bson.NewObjectId().Hex()
-	}
-
 	if performParams.CreateSCMTag {
 		version.Operator = api.APIOperator
 	}
--- a/api/conversion/conversion.go
+++ b/api/conversion/conversion.go
@@ -207,6 +207,8 @@ func ConvertPipelineParamsToVersion(performParams *newapi.PipelinePerformParams)
 	stagesStr := strings.Join(performParams.Stages, ",")
 	version.Operation = api.VersionOperation(strings.Replace(stagesStr, "imageRelease", "publish", 1))
 
+	version.Name = performParams.Name
+
 	if performParams.CreateSCMTag {
 		version.Operator = api.APIOperator
 	}
--- a/api/conversion/conversion.go
+++ b/api/conversion/conversion.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/caicloud/cyclone/api"
 	newapi "github.com/caicloud/cyclone/pkg/api"
-	"gopkg.in/mgo.v2/bson"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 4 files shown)


worker/vcs/manager.go
```diff
@@ -164,6 +165,25 @@ func (vm *Manager) CloneVersionRepository(event *api.Event) error {
 		steplog.InsertStepLog(event, steplog.CloneRepository, steplog.Stop, err)
 		return fmt.Errorf("Unable to clone repository for version: %v\n", err)
 	}
+
+	err = vm.ensureCommitValid(event, destPath, worker)
+	if err != nil {
+		return err
+	}
+
+	vm.formatVersionName(event)
```
supereagle: CloneVersionRepository() just cares about cloning repo, does not need to care about version name. So this logic should not be put here. Please refer to [worker/worker.go#L348-L380](https://github.com/caicloud/cyclone/blob/v1-preview/worker/worker.go#L348-L380).

worker/vcs/manager.go
```diff
@@ -176,23 +196,24 @@ func (vm *Manager) CloneVersionRepository(event *api.Event) error {
...
-			log.Errorf("Unable to push new commit %s :%v\n", event.Version.Commit, err)
-		}
-	}
-	steplog.InsertStepLog(event, steplog.CloneRepository, steplog.Finish, nil)
 	return nil
 }
 
+// formatVersionName replace the random name with default name '$createTime|$commitID' when name empty in create version
+func (vm *Manager) formatVersionName(event *api.Event) {
+	if bson.IsObjectIdHex(event.Version.Name) && event.Version.Commit != "" {
```
supereagle: If the version name is empty, there is no need to assign it an `ObjectIdHex` in the front. Just leave it empty, and assign it the correct value at the right place.

api/conversion/conversion.go
```diff
@@ -208,12 +207,6 @@ func ConvertPipelineParamsToVersion(performParams *newapi.PipelinePerformParams)
 	stagesStr := strings.Join(performParams.Stages, ",")
 	version.Operation = api.VersionOperation(strings.Replace(stagesStr, "imageRelease", "publish", 1))
 
-	if performParams.Name != "" {
-		version.Name = performParams.Name
```
supereagle: The record name provided by user will be missing if remove this.

event/event.go
```diff
@@ -194,7 +193,7 @@ func autoPublishVersion(service *api.Service) {
 		defer ds.Close()
 		version := api.Version{}
 		version.ServiceID = service.ServiceID
-		version.Name = bson.NewObjectId().Hex()
+		version.Name = ""
```
supereagle: Not needed.

worker/worker.go
```diff
@@ -345,6 +347,14 @@ func setEventFailStatus(event *api.Event, ErrorMessage string) {
 	logdog.Error("Operation failed", logdog.Fields{"event": event})
 }
 
+// formatVersionName replace the random name with default name '$createTime|$commitID' when name empty in create version
```
supereagle: Please update the comment.

worker/worker.go
```diff
@@ -345,6 +347,14 @@ func setEventFailStatus(event *api.Event, ErrorMessage string) {
...
+func formatVersionName(event *api.Event) {
```
supereagle: This is a very easy function, does not need to take the complex `*api.Event` as the param, just needs the commit info. What about `getVersionName(commit string) string`? This function will return the version name.

supereagle:
Please fix the comments. As the PR contains many useless commits, you can close this and submit another PR.

PR #256, opened 2017-11-22
fix(#253): use record name as the image tag if not empty

### LINKED ISSUE #253: The tag of built image is not the record name when it is not empty
**What happened**
When specify the record name, the tag of built image is the format commitID-timestamp, which is not the record name.

Perform params:
```json
        {
            "name": "v120",
            "performParams": {
                "stages": [
                    "codeCheckout",
                    "package",
                    "imageBuild",
                    "integrationTest",
                    "imageRelease",
                    "unitTest"
                ]
            },
        }
```

Built image tag: fa4fd28-171121121046

**What you expected to happen**
The tag of built image should be the record name when it is provided.

supereagle:
## Fixes #253

**Changes**
   
  - [x] Use record name as the image tag if not empty

/cc @zz-bmj @LiangXugang

### DIFF
--- a/worker/worker.go
+++ b/worker/worker.go
@@ -357,9 +357,7 @@ func formatVersionName(event *api.Event) {
 
 // setImageNameAndTag sets the image name and tag name of the event.
 func setImageNameAndTag(dockerManager *docker.Manager, event *api.Event) {
-
 	var imageName, tagName string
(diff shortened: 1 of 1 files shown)


zz-bmj:
LGTM

Issue #278, opened 2017-12-06
Check the SCM auth with token when create or update project


supereagle:
**What happened**
SCM auth with username and password is checked when create or update project, but that with token is not checked.

**What you expected to happen**
Check the SCM auth with token when create or update project.

**Reproduce steps** (as minimally and precisely as possible)

PR #280, opened 2017-12-07
fix(#278): check token when creat/update project with token

LiangXugang:
## Fixes [ISSUE-278](https://github.com/caicloud/cyclone/issues/278)

**Changes**
   
  - [x] Check whether the token has authority of repo when create/update project with token
**Reviewers**

  - [ ] @supereagle  please review
    
**Checklist**
   
   - [x] Rebased/mergable
   - [x] Tests pass
   - [x] Issue/task done

### DIFF
--- a/pkg/scm/scm.go
+++ b/pkg/scm/scm.go
@@ -75,11 +75,12 @@ func GenerateSCMToken(config *api.SCMConfig) error {
 	config.Server = strings.TrimSuffix(config.Server, "/")
 
 	scmType := config.Type
-	token := config.Token
 	provider, err := GetSCMProvider(scmType)
 	if err != nil {
 		return err
 	}
+
+	var generatedToken string
 
 	switch scmType {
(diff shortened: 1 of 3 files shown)


pkg/scm/scm.go
```diff
@@ -103,15 +111,18 @@ func GenerateSCMToken(config *api.SCMConfig) error {
...
+			return fmt.Errorf("token is unauthorized to repos")
 		}
 	case api.SVN:
 		return fmt.Errorf("SCM %s is not supported", scmType)
 	default:
 		return fmt.Errorf("SCM type %s is unknow", scmType)
 	}
 
-	// Update the token if generate a new one.
-	config.Token = token
```
supereagle: Add a new variable `generatedToken` for generated token, and check its value here. If it is empty, then check the token of the config, otherwise set the generated token to the config. What about this implementation?

PR #320, opened 2018-01-15
feat(docker): add docker manager

### LINKED ISSUE #319: Refactor Docker manager
Current Docker manager use `event` as the params, which is not easily reused. Need to make it independent of `event`.

supereagle:
## Fixes #319

**Changes**

  - [x] Move the folder into `pkg`
  - [x] Refactor Docker manager to make it independent of `event`

### DIFF
--- a/pkg/docker/client.go
+++ b/pkg/docker/client.go
@@ -1,0 +1,180 @@
+/*
+Copyright 2016 caicloud authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


pkg/docker/client.go
```diff
@@ -0,0 +1,180 @@
...
+			if err != nil {
+				log.Error(err)
+				return err
+			}
+
+			if _, err := io.Copy(fileOutPut, tarReader); err != nil {
+				log.Error(err)
+				return err
+			}
+			fileOutPut.Close()
```
bbbmj: `defer` is better?

supereagle: Have fixed.

PR #337, opened 2018-02-02
feat(#319): add registry auth info into Docker manager

supereagle:

Add registry auth info into Docker manager to support pushing and pulling images from registry.

Fixes partially #319

/cc @bbbmj 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/docker/client.go
+++ b/pkg/docker/client.go
@@ -36,10 +36,12 @@ const (
 // The Docker client can be direclty used for some functions not provided by this manager.
 type DockerManager struct {
 	// Client represets the Docker client.
-	Client *docker_client.Client
+	Client     *docker_client.Client
+	EndPoint   string
+	AuthConfig *docker_client.AuthConfiguration
 }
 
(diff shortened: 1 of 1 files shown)


pkg/docker/client.go
```diff
@@ -36,10 +36,12 @@ const (
...
 type DockerManager struct {
 	// Client represets the Docker client.
-	Client *docker_client.Client
+	Client     *docker_client.Client
+	EndPoint   string
+	AuthConfig *docker_client.AuthConfiguration
 }
 
-func NewDockerManager(endpoint string) (*DockerManager, error) {
+func NewDockerManager(endpoint, registryServer, registryUsername, registryPassword string) (*DockerManager, error) {
```
bbbmj: What's the difference between `endpoint` and `registryServer`?

supereagle: `endpoint` is Docker server endpoint, which is used to new Docker client.
`registryServer` is the registry server.

PR #347, opened 2018-02-16
feat(#346): use cache volume to speed up pipeline

### LINKED ISSUE #346: Speed up pipeline for maven projects

/kind feature

**What happened**:
For maven projects, their dependencies are downloaded for every build. If the network is not good, or the dependency is huge, the pipeline will be very slow.

Cache the dependency to speed up pipeline.

/assign @supereagle
/priority p0

supereagle:

Use cache volume to speed up pipeline

Fixes #346

/cc @bbbmj 

**Release note**:

```release-note
Cache dependency to speed up pipeline.
```

### DIFF
--- a/event/event.go
+++ b/event/event.go
@@ -147,6 +147,7 @@ func createServiceHandler(event *api.Event) error {
 	logdog.Infof("create service handler")
 	opts := workerOptions.DeepCopy()
 	opts.Namespace = getNamespaceFromEvent(event)
+	opts.CacheVolume = getCacheVolumeFromEvent(event)
 	worker, err := CloudController.Provision(string(event.EventID), opts)
 	if err != nil {
 		return err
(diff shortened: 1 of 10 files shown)


event/event.go
```diff
@@ -421,3 +423,19 @@ func getNamespaceFromEvent(event *api.Event) string {
...
 }
+
+// getCacheVolumeFromEvent gets the cache volume from event data for worker. Will return empty string if can not get it.
+func getCacheVolumeFromEvent(event *api.Event) string {
+	if event == nil {
+		log.Error("Can not get cache volume from data as event is nil")
+		return ""
+	}
+
+	if v, e := event.Data["cachevolume"]; e {
```
bbbmj: maybe the map key should be variable?

supereagle: This PR is just for `release-0.3`, it will not be like this in the latest implementation.

supereagle: Done

PR #438, opened 2018-04-16
test(worker):add unit test for ExecPackage

zhujian7:

Add your description

Fixes #

/cc @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/docker/client.go
+++ b/pkg/docker/client.go
@@ -150,7 +150,7 @@ func (dm *DockerManager) StartContainer(options docker_client.CreateContainerOpt
 	}
 
 	// Run the container
-	err = dm.Client.StartContainer(container.ID, nil)
+	err = dm.Client.StartContainer(container.ID, options.HostConfig)
 	if err != nil {
 		return "", fmt.Errorf("start container with error %s", err.Error())
 	}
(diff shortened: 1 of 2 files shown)


pkg/docker/client.go
```diff
@@ -150,7 +150,7 @@ func (dm *DockerManager) StartContainer(options docker_client.CreateContainerOpt
 	}
 
 	// Run the container
-	err = dm.Client.StartContainer(container.ID, options.HostConfig)
+	err = dm.Client.StartContainer(container.ID, nil)
```
supereagle: Why need this change? It will affect [here](https://github.com/caicloud/cyclone/blob/master/pkg/worker/stage/stage.go#L141-L158)

pkg/worker/stage/stage_test.go
```diff
@@ -230,3 +230,114 @@ func TestExecImageBuild(t *testing.T) {
...
+					},
+				},
+			},
+			buildInfo: &api.BuildInfo{
+				BuildTool: &api.BuildTool{
+					Name:    api.MavenBuildTool,
+					Version: "1.0",
+				},
+
+				CacheDependency: true,
```
supereagle: Maybe can not test speedup in unit test, because it depends on volume.

pkg/docker/client.go
zhujian7: can we update the docker in worker to `docker:1.13.1-dind`.
[setting host configuration options when the container starts is deprecated](https://docs.docker.com/engine/deprecated/#hostconfig-at-api-container-start)

PR #567, opened 2018-09-06
fix: change the url path pipelinetemplates to configtemplates

zhujian7:

Add your description

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/api/v1/descriptor/template.go
+++ b/pkg/api/v1/descriptor/template.go
@@ -12,17 +12,17 @@ func init() {
 
 var templates = []definition.Descriptor{
 	{
-		Path:        "/pipelinetemplates",
-		Description: "Template API",
+		Path:        "/configtemplates",
+		Description: "Pipeline Config Template API",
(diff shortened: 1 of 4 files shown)


pkg/api/v1/descriptor/template.go
```diff
@@ -12,17 +12,17 @@ func init() {
...
-		Path:        "/pipelinetemplates",
-		Description: "Template API",
+		Path:        "/configtemplates",
+		Description: "Pipeline Config Template API",
 		Definitions: []definition.Definition{
 			{
 				Method:      definition.List,
 				Function:    handler.ListTemplates,
-				Description: "Get all cyclone built-in pipeline templates",
+				Description: "Get all cyclone built-in pipeline config templates",
```
supereagle: Also need to update description and some comments of function `ListTemplates`.

zhujian7: have fixed

PR #569, opened 2018-09-07
feat: add notification for pipeline

zhujian7:

Add notification for pipeline;
will create a post request hook to configure notification url(`NOTIFICATION_URL` environment) after exec pipeline. 

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -760,13 +760,12 @@ type NotificationContent struct {
 type NotificationPolicyType string
 
 const (
-
 	// AlwaysNotify represents always send notification regardless of the pipeline record result.
-	AlwaysNotify NotificationPolicyType = "always"
+	AlwaysNotify NotificationPolicyType = "Always"
 
 	// SuccessNotify represents send notification when pipeline record execute successfully.
-	SuccessNotify NotificationPolicyType = "success"
+	SuccessNotify NotificationPolicyType = "Success"
 
 	// FailureNotify represents send notification when pipeline record execute failed.
-	FailureNotify NotificationPolicyType = "failure"
+	FailureNotify NotificationPolicyType = "Failure"
 )
--- a/docs/api/v1/api.md
+++ b/docs/api/v1/api.md
@@ -408,7 +408,7 @@ Success:
     "owner": "string", 
     "projectID": "string", 
     "notification": {                 // not implemented
-      "policy": "always",             // string  always,success,failure
+      "policy": "Always",             // string  Always,Success,Failure
       "receivers": [
         {
           "type": "email",            // string  email,webhook,slack


pkg/api/types.go
```diff
@@ -728,3 +742,17 @@ const (
...
+	ProjectName  string    `yaml:"projectName,omitempty" json:"projectName,omitempty" `
+	PipelineName string    `yaml:"pipelineName,omitempty" json:"pipelineName,omitempty" `
+	RecordName   string    `yaml:"recordName,omitempty" json:"recordName,omitempty" `
+	RecordID     string    `yaml:"recordID,omitempty" json:"recordID,omitempty" `
+	Trigger      string    `bson:"trigger,omitempty" json:"trigger,omitempty" description:"trigger of the pipeline record"`
+	Status       Status    `bson:"status,omitempty" json:"status,omitempty" description:"status of the pipeline record"`
+	ErrorMessage string    `bson:"errorMessage,omitempty" json:"errorMessage,omitempty" description:"error message for the pipeline failure"`
+	StartTime    time.Time `bson:"startTime,omitempty" json:"startTime,omitempty" description:"start time of the pipeline record"`
+	EndTime      time.Time `bson:"endTime,omitempty" json:"endTime,omitempty" description:"end time of the pipeline record"`
+	TimeCost     float64   `bson:"timeCost,omitempty" json:"timeCost,omitempty" description:"time cost, unit(second), endTime - startTime"`
```
supereagle: `TimeCost` is not needed to defined, as it can be got by `StartTime` and `EndTime`

pkg/event/manager.go
```diff
@@ -230,6 +235,82 @@ func postHookEvent(event *api.Event) {
...
+
+	// post hooks
+	if pipeline != nil && pipeline.Notification != nil {
+
+		policy := pipeline.Notification.Policy
+
+		sendFlag := false
+		switch record.Status {
+		case api.Success:
+			if policy == "always" || policy == "success" {
```
supereagle: Hard code here is not very good, it would be better to define a `NotificationPolicyType` type.

pkg/event/manager.go
```diff
@@ -230,6 +235,82 @@ func postHookEvent(event *api.Event) {
...
+	pipeline := event.Pipeline
+	record := event.PipelineRecord
+
+	// post hooks
+	if pipeline != nil && pipeline.Notification != nil {
+
+		policy := pipeline.Notification.Policy
+
+		sendFlag := false
+		switch record.Status {
```
supereagle: If `policy` is `always`, there is no need check record status.
So we can first check `policy == "always"`, then check `record.Status` only if `policy != "always"`

PR #571, opened 2018-09-14
chore(*): update retry policy of code checkout

bbbmj:

Update retry policy of code checkout

Fixes #

Reference to #

/cc @supereagle @zhujian7 

**Release note**:

```release-note
Update retry policy of code checkout
```

### DIFF
--- a/pkg/worker/scm/manager.go
+++ b/pkg/worker/scm/manager.go
@@ -35,9 +35,11 @@ const (
 	cloneDir = "/tmp/code"
 
 	repoNameRegexp = `^http[s]?://(?:git[\w]+\.com|[\d]+\.[\d]+\.[\d]+\.[\d]+:[\d]+|localhost:[\d]+)/([\S]*)\.git$`
+)
 
-	// maxRetry represents the max number of retry for git clone.
-	maxRetry = 5
+var (
+	maxRetries   = 8
+	initialDelay = 2 * time.Second
 )
 
 // scmProviders represents the set of SCM providers.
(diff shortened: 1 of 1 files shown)


pkg/worker/scm/manager.go
```diff
@@ -211,14 +213,16 @@ func CloneRepo(token string, codeSource *api.CodeSource, ref string, folder stri
...
 		logs, err = p.Clone(token, url, reference, destPath)
 		if err == nil {
 			return logs, nil
 		}
 
-		logdog.Infof("checkout code as error %v, retry times: %v", err, i)
-		time.Sleep(time.Second * 1)
+		logdog.Infof("failed to checkout, will retry after %d seconds, due to error: %v", backoff, err)
+		time.Sleep(backoff)
+		backoff *= 2
```
zhujian7: What about `backoff += 2` ?
We will wait for almost 512(2 to the 9th power) seconds in the worst case, and i think it is a little long.

bbbmj: It's useless to use `backoff += 2` to retry, in most condition, a service needs more time to recovery from downtime.

supereagle: retry is mainly for unstable environment, not for service recovery.

bbbmj: Just like `pending` in k8s, we give the program a status or a goal, it keeps on trying to make it

supereagle: The `pending` pipelines will hold the resources, then other pipelines will be blocked because lack of resource. 
On the other side, if the pipeline will fail in the end, it is not very good to give the feedback after a long wait.

So, I think retry within a short period is reasonable.

PR #594, opened 2018-11-07
feat: refactoring Cyclone, create k8s crd and generate code

cd1989:

Most of the changes are in the vendor. The effective changes are:

- Clean old code
- Create k8s  crd and generate clients.

Fixes #

Reference to #

/cc @zhujian7 

**Release note**:

```release-note
NONE
```


ddysher:
This is too large to review, please create separate PRs.

Also, do we want to create a new branch?

pkg/apis/cyclone/v1alpha1/stage.go
```diff
@@ -0,0 +1,36 @@
...
+	Template TemplateRef `json:"template,omitempty"`
+}
+
+// StageSpec defines stage specification.
+// Exact one workload should be specified.
+type StageSpec struct {
+	// Pod kind workload
+	Pod PodWorkload `json:"pod,omitempty"`
+	// CRD kind workload
+	CRD CRDWorkload `json:"crd:omitempty"`
```
ddysher: `crd:omitempty` -> `crd,omitempty`

pkg/apis/cyclone/v1alpha1/stage.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// TemplateRef refers to a stage template and defines necessary arguments.
+type TemplateRef struct {
+	// Template name
+	Name string `json:"name"`
```
ddysher: Name is not enough to reference an object, e.g. you can't prevent an object update.  Try use `ObjectReference` in Kubernetes?

pkg/apis/cyclone/v1alpha1/stage_template.go
```diff
@@ -0,0 +1,83 @@
...
+	// Specification of the CRD
+	YML string `json:"yaml"`
+	// How to judge CRD workload is completed
+	StatusRef StatusRef `json:"statusRef"`
+}
+
+// PodWorkload describes pod type workload, a complete pod spec is included.
+type PodWorkload struct {
+	// Stage inputs
+	Inputs Inputs `json:"inputs,omitempty"`
```
ddysher: I think we've made the decision to move this to top level?

pkg/apis/cyclone/v1alpha1/stage_template.go
```diff
@@ -0,0 +1,83 @@
...
+	ComplatedValue string `json:"complatedValue"`
+}
+
+// CRDSpec defines CRD workload specification.
+type CRDSpec struct {
+	// Specification of the CRD
+	YML string `json:"yaml"`
+	// How to judge CRD workload is completed
+	StatusRef StatusRef `json:"statusRef"`
+}
```
ddysher: I think we also need something like PodRef (each input output should have a reference), otherwise how do we know where to mount our input/outputs?

pkg/apis/cyclone/v1alpha1/stage_template.go
```diff
@@ -0,0 +1,83 @@
...
+
+// StageTemplateSpec defines stage template specification.
+// Exact one workload should be specified.
+type StageTemplateSpec struct {
+	// Arguments of the template
+	Arguments []Argument `json:"arguments,omitempty"`
+	// Pod kind workload
+	Pod PodWorkload `json:"pod,omitempty"`
+	// CRD kind workload
+	CRD CRDWorkload `json:"crd:omitempty"`
```
ddysher: hmm, `crd,omitempty`?

pkg/apis/cyclone/v1alpha1/workflow.go
```diff
@@ -0,0 +1,30 @@
...
+type WorkflowSpec struct {
+	Stages StageItem `json:"stageItem"`
+}
+
+// StageItem describes a stage in a workflow.
+type StageItem struct {
+	// Name of stage
+	Name string `json:"name"`
+	// Stages that this stage depends on
+	Depends []string `json:"depends"`
```
ddysher: FYI: a lot of places use stage name as a reference to a stage, but name is really not a unique identifier (e.g. name can change; name can be the same across different namespace).

I think we can stand with this now...

pkg/apis/cyclone/v1alpha1/workflow_param.go
```diff
@@ -0,0 +1,25 @@
...
+	// Metadata for the particular object, including name, namespace, labels, etc
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Workflow param specification
+	Spec WorkflowParamSpec `json:"spec"`
+}
+
+// WorkflowParamSpec describes global parameters a workflow should run with.
+type WorkflowParamSpec struct {
+	// PVC to be used for workflow
+	PVC string `json:"pvc"`
```
ddysher: Is this used for cache purpose? do i have to pass a PVC, what if the parameter is empty?

pkg/apis/cyclone/v1alpha1/workflow_trigger.go
```diff
@@ -0,0 +1,43 @@
...
+
+const (
+	ScheduledTrigger = "Schedule"
+	WebhookTrigger   = "Webhook"
+)
+
+// WorkflowTriggerSpec defines workflow trigger definition.
+type WorkflowTriggerSpec struct {
+	// Name of the workflow to trigger
+	WorkflowRef string `json:"workflowRef"`
```
ddysher: Use objectreference here

```go
// ObjectReference contains enough information to let you inspect or modify the referred object.
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
type ObjectReference struct {
	// +optional
	Kind string
	// +optional
	Namespace string
	// +optional
	Name string
	// +optional
	UID types.UID
	// +optional
	APIVersion string
	// +optional
	ResourceVersion string

	// Optional. If referring to a piece of an object instead of an entire object, this string
	// should contain information to identify the sub-object. For example, if the object
	// reference is to a container within a pod, this would take on a value like:
	// "spec.containers{name}" (where "name" refers to the name of the container that triggered
	// the event) or if no container name is specified "spec.containers[2]" (container with
	// index 2 in this pod). This syntax is chosen only to have some well-defined way of
	// referencing a part of an object.
	// TODO: this design is not final and this field is subject to change in the future.
	// +optional
	FieldPath string
}
```

pkg/apis/cyclone/register.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"github.com/caicloud/cyclone/pkg/common/crd/apis/cyclone/v1alpha1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// GroupName is the group name use in this package
+const GroupName = "cyclone.caicloud.io"
```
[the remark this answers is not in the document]
cd1989: Good point

pkg/apis/cyclone/v1alpha1/resource.go
```diff
@@ -0,0 +1,50 @@
...
+
+// ResourceSpec describes a resource
+type ResourceSpec struct {
+	// Image to resolve this kind of resource.
+	Resolver string `json:"resolver,omitempty"`
+	// Resource type, e.g. image, git, kv, general.
+	Type ResourceType `json:"type"`
+	// Whether persist resource to a dedicated storage (PV).
+	Persistent bool `json:"persistent"`
+	// Whether to pull resource when there already be data
```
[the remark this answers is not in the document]
cd1989: Old data would be removed and pull again.

pkg/apis/cyclone/v1alpha1/stage.go
```diff
@@ -0,0 +1,36 @@
...
+// Stage defines a workflow stage. Only one of Spec and Template should be specified.
+type Stage struct {
+	// Metadata for the resource, like kind and apiversion
+	metav1.TypeMeta `json:",inline"`
+	// Metadata for the particular object, including name, namespace, labels, etc
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Stage specification without using template
+	Spec StageSpec `json:"spec,omitempty"`
+	// Use stage template
+	Template TemplateRef `json:"template,omitempty"`
```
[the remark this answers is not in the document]
cd1989: OK

pkg/apis/cyclone/v1alpha1/stage.go
cd1989: My mistake

pkg/apis/cyclone/v1alpha1/workflow_param.go
cd1989: Yes, this is used for transfer data (e.g. artifacts) within workflow from one stage to another. If no such need, it can be omitted. 

How about letting users decide whether to set it or not.

pkg/apis/cyclone/v1alpha1/resource.go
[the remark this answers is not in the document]
ddysher: this should be clearly documented

pkg/apis/cyclone/v1alpha1/workflow_param.go
ddysher: didn't we already have persistent flag? sounds a bit confusing and duplicated, even though they have different purpose.

cd1989: This PVC is a tmp pvc, it's necessary for the workflow implementation. While the resource persistence is using another dedicated resource PV, resource management (including the PV.) is independent to workflow.  For a workflow, resources (input, output) are different to artifacts.

ddysher: Yeah, I understand the difference. Using PVC here means we give control to our users, but it's fundamentally a temp PVC which means user shouldn't even care about it.

IMO, the API is not consistent with what PVC should do...

PR #618, opened 2018-11-22
PVC GC

cd1989:

After WorkflowRun terminated, we need to clean data on PVC.

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/common/const.go
+++ b/pkg/workflow/common/const.go
@@ -40,6 +40,7 @@ const (
 	WorkflowLabelName         = "cyclone.io/workflow"
 	PodLabelSelector          = "cyclone.io/workflow==true"
 	WorkflowRunAnnotationName = "cyclone.io/workflowrun"
+	GCAnnotationName          = "cyclone.io/gc"
 	StageAnnotationName       = "cyclone.io/stage"
 
 	// Paths in coordinator container.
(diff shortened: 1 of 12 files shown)


pkg/workflow/common/const.go
```diff
@@ -74,3 +75,9 @@ const (
 	// ContainerStateInitialized represents container is Running or Stopped, not Init or Creating.
 	ContainerStateInitialized ContainerState = "Initialized"
 )
+
+// Constants for GC pod.
+const (
+	GCContainerName = "gc"
+	GCDataPath = "/workspace"
```
supereagle: Format

pkg/workflow/controller/handlers/pod/gc.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	return true
+}
+
+// GCPodUpdated handles GC pod update. If GC pod is terminated, it will be deleted.
+func GCPodUpdated(client clientset.Interface, pod *corev1.Pod) {
+	if pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed {
+		if err := client.CoreV1().Pods(pod.Namespace).Delete(pod.Name, &metav1.DeleteOptions{}); err != nil {
+			if r, ok := err.(*errors.StatusError); ok && r.Status().Reason == metav1.StatusReasonNotFound {
```
supereagle: Directly use `errors.IsNotFound(err)`?

cd1989: You are right.

PR #622, opened 2018-11-22
refactor(coordinator): exit immediately when error occurred and check necessity of resources collection 

zhujian7:

refactor coordinator component:
- exit immediately when workload container error
- check whether resources need to be collected
- coordinator support sending events

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```


cmd/workflow/coordinator/main.go
```diff
@@ -38,6 +38,12 @@ func main() {
 	log.Info("Wait workload containers completion ... ")
 	c.WaitWorkloadTerminate()
 
+	// Check if the workload is succeeded.
+	if !c.IsStageSuccess() {
+		log.Errorf("Workload failed.")
```
cd1989: log.Error

cmd/workflow/coordinator/main.go
```diff
@@ -67,8 +73,8 @@ func main() {
 	log.Info("Wait for all other containers completion ... ")
 	c.WaitAllOthersTerminate()
 
-	// Check if the workload is succeeded.
-	if c.IsWorkloadSuccess() {
+	// Check if the workload and resolver containers are succeeded.
+	if c.IsStageSuccess() {
```
cd1989: This is already checked before, no need to check again.

PR #657, opened 2018-12-10
fix: can not exec worker by docker-type cloud

zhujian7:

When running cyclone with docker cloud, the error 
`fail to update event 5c0e4a177014bcadc498fef8 as err: The dotted field 'requests.memory' in 'workerInfo.quota.requests.memory' is not valid for storage.` 
occurred while creating cyclone worker,  and I think we don't need to store quota info in the `Event` 
 Object, so remove it.

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```


supereagle:
> and I think we don't need to store quota info in the `Event`
> Object, so remove it.

Does this work for k8s cloud?

zhujian7:
@supereagle 
> Does this work for k8s cloud?

Yes, of course.

PR #663, opened 2018-12-12
fix: should use new git resource url to create webhook

zhujian7:

Add your description

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/manager/pipeline.go
+++ b/pkg/server/manager/pipeline.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	"github.com/caicloud/nirvana/log"
-	"github.com/zoumo/logdog"
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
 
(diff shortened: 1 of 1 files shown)


pkg/server/manager/pipeline.go
```diff
@@ -386,16 +385,31 @@ func (m *pipelineManager) UpdatePipeline(projectName string, pipelineName string
...
 		scmTrigger := pipeline.AutoTrigger.SCMTrigger
 		if scmTrigger.Webhook != "" {
-			if err := provider.DeleteWebHook(gitSource.Url, scmTrigger.Webhook); err != nil {
+			if err := provider.DeleteWebHook(oldGitSource.Url, scmTrigger.Webhook); err != nil {
 				return nil, err
 			}
 		}
 	}
 
+	var newGitSource *api.GitSource
```
supereagle: Create webhook for new repo before delete old webhook, then can directly return if there is error.
This can avoid the risk of failing to recreate old webhook later.

zhujian7: If new URL equals the old one, new webhook will be deleted.

PR #665, opened 2018-12-14
feat(*): add cyclone web base code

qiuxiaolian:

Add your description

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```


client/.gitignore
```diff
@@ -0,0 +1,22 @@
...
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+.idea
```
artecher: also add `.vscode` ?

client/package.json
```diff
@@ -0,0 +1,117 @@
...
+  "name": "react-web-demo",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "add": "^2.0.6",
+    "ajv": "^5.5.2",
+    "antd": "^3.0.3",
+    "autoprefixer": "7.1.6",
+    "axios": "^0.18.0",
+    "babel-core": "6.26.0",
```
cd1989: Why not `devDependencies` for babel ?

client/src/components/pipeline/component/list.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
```
artecher: about code convention, how about make all js quote a `'`. eg. `import React from 'react'`, this is more common in js project

client/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "React App",
```
cd1989: How about change to Cyclone ?

client/src/api/http.js
```diff
@@ -0,0 +1,149 @@
...
+// getToken
+const getToken = () => {
+  if (localStorage.getItem("user")) {
+    return JSON.parse(localStorage.getItem("user")).accessToken;
+  }
+  return "";
+};
+
+// hint
+let curMes = "";
```
cd1989: `msg` for `message`

client/src/api/http.js
```diff
@@ -0,0 +1,149 @@
...
+      setTimeout(function() {
+        _message(error.response.data.message);
+      }, 300);
+    }
+
+    return Promise.reject(error);
+  }
+);
+
+//重写instance.get
```
cd1989: Comments should try to explain why we need to rewrite `instance.get`, otherwise the comment is meaningless.

client/src/components/pipeline/component/list.js
cd1989: I'd suggest this as our style conventions: https://github.com/airbnb/javascript

client/package.json
```diff
@@ -0,0 +1,120 @@
...
+  "name": "react-web-demo",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "axios": "^0.18.0",
+    "chalk": "1.1.3",
+    "dotenv": "4.0.0",
+    "fs-extra": "3.0.1",
+    "lodash": "^4.17.11",
+    "mockjs": "^1.0.1-beta3",
```
artecher: Are `mockjs` and `yarn`  needed in dependencies?

client/package.json
```diff
@@ -0,0 +1,120 @@
...
+    "eslint-plugin-jsx-a11y": "5.1.1",
+    "eslint-plugin-prettier": "^3.0.0",
+    "eslint-plugin-react": "7.4.0",
+    "extract-text-webpack-plugin": "3.0.2",
+    "file-loader": "1.1.5",
+    "html-webpack-plugin": "2.29.0",
+    "jest": "20.0.4",
+    "less": "^2.7.3",
+    "less-loader": "^4.0.5",
+    "mobx": "^3.4.1",
```
artecher: Should `mobx` be in dependencies?
I would suggest a test after you move all the lib around.

client/package.json
```diff
@@ -101,17 +71,50 @@
...
     ]
   },
-  "eslintConfig": {
-    "extends": "react-app",
-    "globals": {
-      "_": true
-    }
-  },
   "devDependencies": {
+    "antd": "^3.0.3",
```
cd1989: I don't think this should go here.

client/package.json
cd1989: I think it should be in dependencies since it's similar to redux.

caosm: I think `dependencies` is not necessary, everything would be bundled into static-path before runtime

client/src/api/http.js
```diff
@@ -0,0 +1,140 @@
...
+  setTimeout(function() {
+    loadingNum--;
+    if (loading && loadingNum <= 0) {
+      loading();
+    }
+  }, 0);
+};
+// axios instance
+const instance = axios.create({
+  baseURL: 'http://192.168.19.96:30020/api/v1',
```
caosm: use env instead?

client/src/api/index.js
```diff
@@ -0,0 +1,14 @@
+import http from './http.js';
+
+console.log('instance', http);
```
caosm: remove

PR #695, opened 2018-12-29
fix: fix bug of release chart

codeflitting:

* default_resource_quota： Without default_resource_quota can't start pod
* native-cyclone-server：This causes the pv name to be created too long

Fixes #

Reference to #

/cc @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/release/workflow-controller.yaml
+++ b/release/workflow-controller.yaml
@@ -42,6 +42,16 @@ _config:
             "limits": {
               "max_workflowruns": 50
             },
+            "default_resource_quota": {
+              "limits": {
+                "cpu": "200m",
+                "memory": "256Mi"
+              },
+              "requests": {
+                "cpu": "100m",
+                "memory": "128Mi"
+              }
+            },
             "pvc": "native-cyclone-server-server-v1-0-cyclone-data",
-            "cyclone_server_addr": "native-cyclone-server.default:7099"
+            "cyclone_server_addr": "native-cyclone-server.default:7089"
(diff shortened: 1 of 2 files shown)


release/workflow-controller.yaml
```diff
@@ -42,6 +42,12 @@ _config:
             "limits": {
               "max_workflowruns": 50
             },
+            "default_resource_quota": {
```
supereagle: Missing limit? Please refer to https://github.com/caicloud/cyclone/blob/develop/manifests/cyclone.yaml#L29-L37

codeflitting: need limit？

supereagle: yes

release/workflow-controller.yaml
```diff
@@ -42,6 +42,12 @@ _config:
...
+              "requests": {
+                "cpu": "100m",
+                "memory": "128Mi"
+              }
+            },
             "pvc": "native-cyclone-server-server-v1-0-cyclone-data",
             "cyclone_server_addr": "native-cyclone-server.default:7099"
```
supereagle: Please change the port to `7089`, I change this by mistake yesterday.

codeflitting: done

release/workflow-controller.yaml
codeflitting: done

PR #700, opened 2019-01-02
fix(trigger): reduce the name length of wfr created by trigger

tskdsb:

修复issue：https://github.com/caicloud/clever-admin/issues/256
缩短由trigger创建的wfr name

Fixes #

Reference to #

/assign @cd1989 
/cc @codeflitting 

**Release note**:

```release-note
NONE
```


pkg/workflow/controller/handlers/workflowtrigger/cron_job.go
```diff
@@ -91,7 +92,15 @@ func ToWorkflowTrigger(obj interface{}) (*v1alpha1.WorkflowTrigger, error) {
 
 func (c *CronTrigger) Run() {
 
-	c.WorkflowRun.Name = c.WorkflowTriggerName + "-" + strings.Replace(uuid.NewV1().String(), "-", "", -1)
+	var wfrName string
+	for {
+		wfrName = names.SimpleNameGenerator.GenerateName(c.WorkflowTriggerName + "-")
+		_, err := c.Manage.Client.CycloneV1alpha1().WorkflowRuns(c.Namespace).Get(wfrName, v1.GetOptions{})
```
cd1989: If there is problem to request API server, it would loop infinitely here.

cd1989: In my opinion, we only retry when the error is name already exist error.

Issue #671, opened 2018-12-18
Integration Center


cd1989:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

### Function Details

**Integration Center** manages external services such as SCM(Git, SVN), Docker registry.
Each service has one type: `git`, `svn`, `registry`. For some service type, there can be subTypes, for example, `git` can have subType `gitlab`, `github`.

This function should support:

- CRUD of integrated service 

**What happened**:

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

supereagle:
Have done

/close

PR #736, opened 2019-01-18
feat(api): add api for integration

zhujian7:
- integration CRUD

Add your description

Fixes #

Reference to #671 

/cc @your-reviewer

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/apis/v1alpha1/types.go
+++ b/pkg/server/apis/v1alpha1/types.go
@@ -29,7 +29,7 @@ type TenantSpec struct {
 	PersistentVolumeClaim PersistentVolumeClaim `json:"persistentVolumeClaim"`
 
 	// ResourceQuota describes the resource quota of the namespace,
-	// eg map[string]string{"cpu": "2Gi", "memory": "512Mi"}
+	// eg map[core_v1.ResourceName]string{"cpu": "2Gi", "memory": "512Mi"}
 	ResourceQuota map[core_v1.ResourceName]string `json:"resourceQuota"`
 }
 
@@ -40,7 +40,7 @@ type PersistentVolumeClaim struct {
 	// Name string `json:"name"`
 
 	// StorageClass represents the strorageclass used to create pvc
-	StorageClass string `json:"storageclass"`
+	StorageClass string `json:"storageClass"`
 
 	// Size represents the capacity of the pvc, unit supports 'Gi' or 'Mi'
 	// More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#capacity
@@ -67,6 +67,8 @@ const (
 	SCM IntegrationType = "SCM"
 	// Cluster is the Cluster integration
 	Cluster IntegrationType = "Cluster"
+	// GeneralIntegration is the General integration
+	GeneralIntegration IntegrationType = "General"
 )
 
 // IntegrationSpec contains the integration spec information
@@ -93,6 +95,17 @@ type IntegrationSource struct {
 	// Users can define which cluster will be used to run workload,
 	// and clusters integrated here can be used to deploy application in CD tasks.
 	Cluster *ClusterSource `json:"cluster"`
+
+	// General contains parameters defined by users.
+	General []ParameterItem `json:"general"`
+}
+
+// ParameterItem defines a parameter
+type ParameterItem struct {
+	// Name of the parameter
+	Name string `json:"name"`
+	// Value of the parameter
+	Value string `json:"value"`
 }
 
 // SonarQubeSource represents a code scanning tool for CI.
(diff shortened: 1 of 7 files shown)


pkg/server/apis/v1alpha1/types.go
```diff
@@ -2,23 +2,24 @@ package v1alpha1
...
+	Metadata Metadata `json:"metadata"`
 	// Spec contains tenant spec
 	Spec TenantSpec `json:"spec"`
 }
 
-// TenantMetadata contains metadata information about tenant
-type TenantMetadata struct {
-	// Name is the name of tenant, unique.
+// Metadata contains metadata information
+type Metadata struct {
```
supereagle: What about use k8s metadata?

pkg/server/common/cluster.go
```diff
@@ -0,0 +1,253 @@
...
+// CreateNamespace creates a namespace
+func CreateNamespace(name string, client *kubernetes.Clientset) error {
+	namespace, err := buildNamespace(name)
+	if err != nil {
+		log.Warningf("Build namespace %s error %v", name, err)
+		return err
+	}
+
+	_, err = client.CoreV1().Namespaces().Create(namespace)
+	if err != nil {
```
supereagle:
```suggestion
	if _, err = client.CoreV1().Namespaces().Create(namespace); err != nil {
```

pkg/server/handler/v1alpha1/tenant.go
```diff
@@ -3,26 +3,26 @@ package v1alpha1
...
 
 	"github.com/caicloud/nirvana/log"
 	"k8s.io/api/core/v1"
 	core_v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/util/retry"
 
-	apiv1alpha1 "github.com/caicloud/cyclone/pkg/server/apis/v1alpha1"
+	api "github.com/caicloud/cyclone/pkg/server/apis/v1alpha1"
```
supereagle: Do not recommend this change, if there is another version of API, we still need to specify the version.

pkg/server/handler/v1alpha1/tenant.go
```diff
@@ -117,49 +178,64 @@ func CreateAdminTenant() error {
...
 	}
 
-	return createCycloneTenant(tenant)
+	return createTenant(tenant)
+}
+
+func createControlClusterIntegration(tenant string) error {
+	in := &api.Integration{
+		Metadata: api.Metadata{
+			Name:        common.ControlClusterName,
```
supereagle: I think there is no need to create control cluster, use can directly in default.

pkg/server/handler/v1alpha1/tenant.go
zhujian7: `tenant.go` is under the `v1alpha1` folder, so it is less likely it needs another version of API.

pkg/server/handler/v1alpha1/tenant.go
zhujian7: I don't think so, as control cluster could be closed by users, so it is necessary to hold an instance(secret) at the backend.

pkg/server/apis/v1alpha1/types.go
zhujian7: I think there is a litter different between them:
- no `Description` field in k8s metadata.
- some field in k8s metadata we don't need, e.g. `UID` `ResourceVersion` `Generation` `Initializers` etc.

pkg/server/apis/v1alpha1/types.go
```diff
@@ -45,3 +46,144 @@ type PersistentVolumeClaim struct {
...
+	// DockerRegistry describes info about external system docker registry, and is used to manager containers.
+	DockerRegistry *DockerRegistrySource `json:"dockerRegistry"`
+
+	// SCM describes info about external Source Code Management system, and is used to manager code.
+	SCM *SCMSource `json:"scm"`
+
+	// Cluster contains information about clusters.
+	// Users can define which cluster will be used to run workload,
+	// and clusters integrated here can be used to deploy application in CD tasks.
+	Cluster *ClusterSource `json:"cluster"`
```
cd1989: We can add one more general type to extend it.

pkg/server/apis/v1alpha1/types.go
```diff
@@ -45,3 +46,144 @@ type PersistentVolumeClaim struct {
...
+	Token string `json:"token"`
+}
+
+// ClusterSource contains info about clusters.
+type ClusterSource struct {
+	// Credential is the credential info of the cluster
+	Credential ClusterCredential `json:"credential"`
+	// InCluster describes whether the cluster is the control cluster itself
+	InCluster bool `json:"inCluster,omitempty"`
+	// Worker defines whether this cluster can be used to perform workload.
```
cd1989: `perform workload` --> `run workflow`

pkg/server/apis/v1alpha1/types.go
```diff
@@ -45,3 +46,144 @@ type PersistentVolumeClaim struct {
...
+	// Token is the credential to access SCM.
+	Token string `json:"token"`
+}
+
+// ClusterSource contains info about clusters.
+type ClusterSource struct {
+	// Credential is the credential info of the cluster
+	Credential ClusterCredential `json:"credential"`
+	// InCluster describes whether the cluster is the control cluster itself
+	InCluster bool `json:"inCluster,omitempty"`
```
cd1989: `InCluster` -> `IsControlCluster`

pkg/server/apis/v1alpha1/types.go
```diff
@@ -45,3 +46,144 @@ type PersistentVolumeClaim struct {
...
+	// Worker defines whether this cluster can be used to perform workload.
+	// True, will create namespace and pvc associated with tenant in the cluster.
+	// False, will delete namespace and pvc associated with tenant in the cluster.
+	Worker bool `json:"worker"`
+	// Namespace is the namespace where workload will run in. and if this is not nil, cyclone will
+	// use this namespace and not to create another one.
+	// Useful when 'Worker' is True.
```
cd1989: // Namespace is the namespace where workflows run in. If set, cyclone will use it directly, otherwise a new one will be created.
// It's used when 'Worker' is True.

pkg/server/common/cluster.go
cd1989: Do we need to ignore the namespace already exist error ?

pkg/server/common/cluster.go
```diff
@@ -0,0 +1,253 @@
...
+		log.Errorf("Create namespace %s error %v", name, err)
+		return err
+	}
+
+	return nil
+}
+
+func buildNamespace(name string) (*core_v1.Namespace, error) {
```
cd1989: I suggest to use `tenant` as parameter name.

pkg/server/apis/v1alpha1/types.go
```diff
@@ -45,3 +46,144 @@ type PersistentVolumeClaim struct {
...
+// ClusterSource contains info about clusters.
+type ClusterSource struct {
+	// Credential is the credential info of the cluster
+	Credential ClusterCredential `json:"credential"`
+	// InCluster describes whether the cluster is the control cluster itself
+	InCluster bool `json:"inCluster,omitempty"`
+	// Worker defines whether this cluster can be used to perform workload.
+	// True, will create namespace and pvc associated with tenant in the cluster.
+	// False, will delete namespace and pvc associated with tenant in the cluster.
+	Worker bool `json:"worker"`
```
cd1989: `IsWorkerCluster`

zhujian7:
@cd1989 PTAL

PR #746, opened 2019-01-23
fix(*): fix mobx does not contain an export named useStrict error

qiuxiaolian:

screen shot 2019-01-23 at 6 43 07 pm

Fixes #

Reference to #

/cc @caosm 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/web/src/less/inc.var.less
+++ b/web/src/less/inc.var.less
@@ -63,3 +63,22 @@ code {
     }
   }
 }
+
+.cyclone-layout-header {
+  z-index: 200;
+  width: 100%;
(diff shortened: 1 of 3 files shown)


web/src/less/inc.var.less
```diff
@@ -63,3 +63,31 @@ code {
     }
   }
 }
+
+.cyclone-layout-header {
```
caosm: duplicate?

qiuxiaolian: duplicate from index.less, keep the index.less  only for introducing other files

PR #776, opened 2019-01-31
fix: display project alias instead of name

cd1989:

Display project alias instead of name.

image

Fixes #

Reference to #

/cc @supereagle 

**Release note**:

```release-note
NONE
```


web/src/components/project/component/List.js
```diff
@@ -44,7 +40,7 @@ class List extends React.Component {
     const columns = [
       {
         title: intl.get('name'),
-        dataIndex: 'metadata.name',
+        dataIndex: 'metadata.annotations["cyclone.io/alias"]',
```
huxiangtao: if get undefined , loadash more better?

cd1989: Here `metadata.annotations["cyclone.io/alias"]` is key path in the object, not retrieve data here.

PR #786, opened 2019-02-11
chore(integrationCenter):modify integration data source type key 

huxiangtao:

replace integration data source type key to Camel-Case format

Fixes #

Reference to #

/cc @caosm @cd1989 @qiuxiaolian 

**Release note**:

```release-note
NONE
```


caosm:
`chore` or `fix` would be more apropos than `feat`

cd1989:
> `chore` or `fix` would be more apropos than `feat`

Please also change the commit message @huxiangtao

PR #787, opened 2019-02-13
refactor(server): configure cyclone server by configmap

zhujian7:

configure cyclone server by configmap instead of flags

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/caicloud/cyclone/pkg/server/handler"
 	"github.com/caicloud/cyclone/pkg/server/handler/v1alpha1"
 	"github.com/caicloud/cyclone/pkg/server/version"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/caicloud/cyclone/pkg/server/biz/tenants"
 	"github.com/caicloud/nirvana"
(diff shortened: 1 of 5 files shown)


cmd/server/main.go
```diff
@@ -39,34 +40,28 @@ import (
 
 // APIServerOptions contains all options(config) for api server
 type APIServerOptions struct {
```
cd1989: I'd prefer name `Options` as it's in `server` package.

cmd/server/main.go
```diff
@@ -39,34 +40,28 @@ import (
...
-	CycloneAddr string
-
-	Loglevel string
-
-	// StorageClass is used to create pvc for default tenant
-	StorageClass string
+	// ConfigMap that configures for cyclone server
+	ConfigMap string
+	// Namespace that cyclone server will run in
+	Namespace string
```
cd1989: I think these two items should have default values, please describe them here.

pkg/server/config/config.go
```diff
@@ -1,130 +1,83 @@
...
+	if config.CycloneServerHost == "" {
+		log.Warningf("CycloneServerHost not configured, cyclone server can not listen API requests.")
 	}
 
-	intVal, err := strconv.ParseInt(val, 10, 32)
-	if err != nil {
-		log.Fatalf("The value '%s' of environment variable '%s' is not a valid integer: %v\n", val, name, err)
-		os.Exit(-1)
+	if config.CycloneServerPort == 0 {
+		log.Warningf("CycloneServerPort not configured, cyclone server can not listen API requests.")
```
cd1989: If `cyclone server can not listen API requests`, should return `false` in my understand. And I think here we should fallback to default value. The same to server host, these two config items should be optional for users.

zhujian7: OK

zhujian7:
@cd1989  All comments have been fixed, PTAL

PR #790, opened 2019-02-15
fix: return success if ns,quota,pvc exist while open worker cluster

zhujian7:
- return success if ns,quota,pvc exist while open worker cluster
- return a specific error to indicate creating control cluster integration error while creating tenant
- do not open the control cluster to run workflow while creating tenant
- some refactor about give default value to cyclone server config

Fixes #

Reference to #

/cc @cd1989 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/config/config.go
+++ b/pkg/server/config/config.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/caicloud/nirvana/log"
 	core_v1 "k8s.io/api/core/v1"
+
+	"github.com/caicloud/cyclone/pkg/server/common"
 )
 
 const (
@@ -65,12 +67,19 @@ func LoadConfig(cm *core_v1.ConfigMap) error {
 		return fmt.Errorf("validate config failed")
 	}
 
+	modifier(&Config)
+
 	log.Infof("cyclone server config: %v", Config)
 	return nil
 }
 
 // validate validates some required configurations.
 func validate(config *CycloneServerConfig) bool {
+	return true
+}
+
+// modifier modifies the config, give the config some default value if they are not set.
+func modifier(config *CycloneServerConfig) {
 	if config.CycloneServerHost == "" {
 		log.Warningf("CycloneServerHost not configured, will use default value '0.0.0.0'")
 		config.CycloneServerHost = "0.0.0.0"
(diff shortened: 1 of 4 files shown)


pkg/server/config/config.go
```diff
@@ -65,12 +67,19 @@ func LoadConfig(cm *core_v1.ConfigMap) error {
...
 
+	modifier(&Config)
+
 	log.Infof("cyclone server config: %v", Config)
 	return nil
 }
 
 // validate validates some required configurations.
 func validate(config *CycloneServerConfig) bool {
+	return true
```
supereagle: Always return true?

pkg/server/config/config.go
```diff
@@ -81,5 +90,19 @@ func validate(config *CycloneServerConfig) bool {
...
 	}
 
-	return true
+	if config.DefaultPVCConfig.Size == "" {
+		log.Warningf("DefaultPVCConfig.Size not configured, will use default value '5Gi'")
+		config.DefaultPVCConfig.Size = common.DefaultPVCSize
+	}
+
+	if config.WorkerNamespaceQuota == nil {
+		log.Warningf("WorkerNamespaceQuota not configured, will use default value limits.cpu limits.memory requests.cpu requests.memory")
```
supereagle: What about `WorkerNamespaceQuota not configured, will use default quota`?

supereagle: Use `Warningln` instead of `Warningf`, as there is no format.

supereagle: Use `Warningln` instead of `Warningf`, as there is no format.

pkg/server/config/config.go
zhujian7: There is nothing we need to validate by now, I think I need to delete this function at this time.

pkg/server/config/config.go
zhujian7: Ok, i will use `Warningln`

zhujian7: What do you mean `default quota`? empty?

zhujian7: Ok, i will use `Warningln`

supereagle: > What do you mean `default quota`? empty?

Just change the log content to make it simpler.

pkg/server/config/config.go
supereagle: ok

zhujian7:
@supereagle PTAL

PR #823, opened 2019-03-08
feat: add workflow label for workflowrun

zhujian7:

Add workflow label for workflowrun to restrict listing workflowrun

Fixes #

Reference to #

/cc @cd1989 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/common/const.go
+++ b/pkg/workflow/common/const.go
@@ -63,8 +63,8 @@ const (
 
 	// WorkflowLabelName is label to indicate resources created by Cyclone workflow engine
 	WorkflowLabelName = "cyclone.io/workflow"
-	// WorkflowRunLabelName is label applied to WorkflowRun to specify Workflow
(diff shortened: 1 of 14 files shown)


pkg/workflow/common/const.go
```diff
@@ -63,8 +63,8 @@ const (
 
 	// WorkflowLabelName is label to indicate resources created by Cyclone workflow engine
 	WorkflowLabelName = "cyclone.io/workflow"
-	// WorkflowRunLabelName is label applied to WorkflowRun to specify Workflow
-	WorkflowRunLabelName = "cyclone.io/workflow-name"
+	// WorkflowNameLabelName is label applied to WorkflowRun to specify Workflow
+	WorkflowNameLabelName = "cyclone.io/workflow-name"
```
cd1989: I'd suggest to extract these label const to `pkg/common`, so that cyclone server and workflow engine can share them. You can create a new PR for it.

PR #842, opened 2019-03-16
feat(#841): Send notifications after workflowruns finish

### LINKED ISSUE #841: Send notifications after workflowruns finish

/kind feature

**What happened**:

Send notifications after workflowruns finish to easily integrate with other systems.

supereagle:

Add your description

Fixes #841

Reference to #

/cc @cd1989  @zhujian7 

**Release note**:

```release-note
NONE
```


pkg/server/handler/v1alpha1/workflow.go
```diff
@@ -20,6 +20,7 @@ func CreateWorkflow(ctx context.Context, project, tenant string, wf *v1alpha1.Wo
 	for _, modifier := range modifiers {
 		err := modifier(tenant, project, "", wf)
 		if err != nil {
+			log.Infof("Failed to create workflow %s for tenant %s in project %s as error: %v", wf.Name, tenant, project, err)
```
zhujian7: Infof -> Errorf?

pkg/workflow/controller/handlers/workflowrun/handler.go
```diff
@@ -84,11 +104,45 @@ func (h *Handler) ObjectUpdated(obj interface{}) {
...
 	if originWfr.Status.Overall.Status == v1alpha1.StatusCompleted ||
 		originWfr.Status.Overall.Status == v1alpha1.StatusError ||
-		originWfr.Status.Overall.Status == v1alpha1.StatusWaiting ||
 		originWfr.Status.Overall.Status == v1alpha1.StatusCancelled {
+		// No need to send notifications for workflowruns finished before workflow controller starts.
+		if originWfr.Status.Overall.LastTransitionTime.Before(controllerStartTime) {
+			return
+		}
+
+		// Send notifications with workflowrun.
```
zhujian7: Could you please extract this *sending notification code block* to a method ?

pkg/workflow/controller/config.go
```diff
@@ -98,6 +100,17 @@ type LimitsConfig struct {
 	MaxWorkflowRuns int `json:"max_workflowruns"`
 }
 
+// NotificationEndpoint represents the config of notification endpoint.
+// Workflow controller will send notifications after workflowruns finish
+// if notification endpoints are configured.
+type NotificationEndpoint struct {
+	// Name represents the name of notification endpoint.
+	Name string `json:"name"`
```
zhujian7: Where do we need this `Name` ?

supereagle: We support multiple notification endpoints, need a name to easily distinguish them.

pkg/workflow/controller/config.go
```diff
@@ -61,6 +61,8 @@ type WorkflowControllerConfig struct {
 	Secret string `json:"secret"`
 	// CycloneServerAddr is address of the Cyclone Server
 	CycloneServerAddr string `json:"cyclone_server_addr"`
+	// Notifications represents the config to send notifications after workflowruns finish.
+	Notifications []NotificationEndpoint `json:"notifications"`
```
cd1989: What kinds of notification are configured here ?

pkg/workflow/controller/config.go
cd1989: We can't distinguish them by URL ? With name, users have to guarantee uniqueness of names.

pkg/workflow/controller/handlers/workflowrun/handler.go
```diff
@@ -84,11 +104,45 @@ func (h *Handler) ObjectUpdated(obj interface{}) {
...
+		bodyBytes, err := json.Marshal(originWfr)
```
cd1989: I'd like a field in status of wfr to mark whether notification been sent and the send result(send successfully, or failed). This should be a status of wfr.

And I think there is no need to send the whole `wfr`. We need define the notification request as need.

pkg/workflow/controller/config.go
supereagle: yes

pkg/workflow/controller/config.go
supereagle: We do not depend on the name, is for users. We can make this as optional.

pkg/workflow/controller/handlers/workflowrun/handler.go
supereagle: > I'd like a field in status of wfr to mark whether notification been sent and the send result(send successfully, or failed). This should be a status of wfr.

This can be added as an enhancement later.

> And I think there is no need to send the whole `wfr`. We need define the notification request as need.

Sending the whole `wfr` to keep the extensibility, implementers can get what they need from `wfr`.

pkg/workflow/controller/handlers/workflowrun/handler.go
supereagle: Done

pkg/server/handler/v1alpha1/workflow.go
supereagle: Done

pkg/workflow/controller/handlers/workflowrun/handler.go
cd1989: > Sending the whole wfr to keep the extensibility, implementers can get what they need from wfr.

I hold different opinions against this, couple the API spec to the definition of wfr is not a good pattern. wfr is easier to be modifed (add new fields, rename fields), but API spec should consider compatibility.

pkg/workflow/controller/config.go
cd1989: So name is just a display name for users to read ?

pkg/workflow/controller/handlers/workflowrun/handler.go
supereagle: wfr is a part of API spec as there is CRUD API for it, we should also consider its compatibility.

cd1989: Notification and wfr CRUD are two different API sets, they should not bound together. If we have to break the compatibility of wfr one day, it wont affect notifications.

supereagle: Have discussed offline, and agreed with:

Notification should include basic informations such as tenant, project, etc, but workflow engine does not aware about these concepts. Workflow engine can not provide these necessary informations, so we directly send `wfr` in notification and let implementers get what they need from it.

PR #846, opened 2019-03-18
feat(server): add api to list all pods for one tenant

supereagle:

Add api to list all working pods for one tenant to let users knows about the workload of their workflows.

Fixes #

Reference to #

/cc @zhujian7 @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/common/const.go
+++ b/pkg/server/common/const.go
@@ -77,4 +77,9 @@ const (
 
 	// ControlClusterName is the name of control cluster
 	ControlClusterName = "control-cluster"
+
+	// PodLabelSelector is selector used to select pod created by Cyclone stages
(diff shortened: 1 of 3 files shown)


pkg/server/common/const.go
```diff
@@ -77,4 +77,9 @@ const (
 
 	// ControlClusterName is the name of control cluster
 	ControlClusterName = "control-cluster"
+
+	// PodLabelSelector is selector used to select pod created by Cyclone stages
+	// TODO(robin) Copy from pkg/workflow/common/const.go, need to create package pkg/common
```
cd1989: How about resolve it in this PR.

supereagle: Will do this in another PR #848

PR #863, opened 2019-03-25
feat(owner): support owner and trigger

supereagle:

Add your description

Fixes #

Reference to #

/cc @zhujian7 @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/common/const.go
+++ b/pkg/server/common/const.go
@@ -75,6 +75,9 @@ const (
 	// AnnotationTrigger is the annotation key used to indicate the trigger of workflowruns.
 	AnnotationTrigger = "cyclone.io/trigger"
 
+	// CronTimerTrigger represents the trigger of workflowruns triggered by cron timer.
+	CronTimerTrigger = "cron-timer"
+
 	// QuotaCPULimit represents default value of 'limits.cpu'
 	QuotaCPULimit = "2"
 	// QuotaCPURequest represents default value of 'requests.cpu'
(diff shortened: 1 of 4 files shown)


pkg/server/common/const.go
```diff
@@ -69,6 +69,12 @@ const (
 	// AnnotationStorageUsage is annotation to store storage usuage information
 	AnnotationStorageUsage = "storage.cyclone.io/usage"
 
+	// AnnotationOwner is the annotation key used to indicate the owner of resources.
+	AnnotationOwner = "cyclone.io/owner"
```
cd1989: I suggest to use `Creator` instead of `Owner`, otherwise it would confuse with owner of resources in k8s.

pkg/server/handler/v1alpha1/webhook.go
```diff
@@ -143,6 +143,9 @@ func createWorkflowRun(tenant, wftName string, data *scm.EventData) error {
 	wfr := &v1alpha1.WorkflowRun{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: tag,
+			Annotations: map[string]string{
+				common.AnnotationTrigger: string(data.Type),
```
cd1989: For other types of wfr, still need trigger annotation, IMO.

pkg/server/common/const.go
supereagle: If you support user system later, users can change the owner, but the creator is fixed, not suitable to change.

pkg/server/handler/v1alpha1/webhook.go
supereagle: Have added trigger for workflowruns triggered by cron timer

PR #865, opened 2019-03-26
feat: disable wfr acceleration when pvc usage exceed limit

cd1989:

Disable acceleration when PVC usage exceed limit.

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/biz/usage/usage.go
+++ b/pkg/server/biz/usage/usage.go
@@ -1,0 +1,171 @@
+package usage
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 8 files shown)


pkg/server/biz/usage/usage.go
```diff
@@ -0,0 +1,172 @@
...
+	UsedPercentage(folder string) (float64, error)
+}
+
+type pvcReporter struct {
+	tenant string
+	usage  *PVCUsageFloat64
+}
+
+// NewPVCReporter creates a PVC usage reporter.
+func NewPVCReporter(client clientset.Interface, tenant string) (PVCReporter, error) {
```
zhujian7: param `client` is not used ?

PR #866, opened 2019-03-26
feat(api): support search and filter

supereagle:

Add your description

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/admin.go
+++ b/pkg/server/handler/v1alpha1/admin.go
@@ -13,7 +13,7 @@ import (
 )
 
 // AllWorkflows lists all workflows matched the provided label from all projects and tenants.
-func AllWorkflows(ctx context.Context, label string, pagination *types.Pagination) (*types.ListResponse, error) {
+func AllWorkflows(ctx context.Context, label string, query *types.QueryParams) (*types.ListResponse, error) {
 	workflows, err := handler.K8sClient.CycloneV1alpha1().Workflows("").List(metav1.ListOptions{
 		LabelSelector: label,
 	})
@@ -24,20 +24,20 @@ func AllWorkflows(ctx context.Context, label string, pagination *types.Paginatio
 
 	items := workflows.Items
 	size := int64(len(items))
-	if pagination.Start >= size {
+	if query.Start >= size {
 		return types.NewListResponse(int(size), []v1alpha1.Workflow{}), nil
 	}
 
-	end := pagination.Start + pagination.Limit
+	end := query.Start + query.Limit
 	if end > size {
 		end = size
 	}
 
-	return types.NewListResponse(int(size), items[pagination.Start:end]), nil
+	return types.NewListResponse(int(size), items[query.Start:end]), nil
 }
 
 // AllStages lists all stages matched the provided label from all projects and tenants.
-func AllStages(ctx context.Context, label string, pagination *types.Pagination) (*types.ListResponse, error) {
+func AllStages(ctx context.Context, label string, query *types.QueryParams) (*types.ListResponse, error) {
 	stages, err := handler.K8sClient.CycloneV1alpha1().Stages("").List(metav1.ListOptions{
 		LabelSelector: label,
 	})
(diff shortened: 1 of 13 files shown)


pkg/server/handler/v1alpha1/admin.go
```diff
@@ -24,20 +24,20 @@ func AllWorkflows(ctx context.Context, label string, pagination *types.Paginatio
 
 	items := workflows.Items
 	size := int64(len(items))
-	if pagination.Start >= size {
+	if qp.Start >= size {
```
cd1989: I suggest to use `query`, `qp` is unreadable.

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -55,7 +57,35 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
...
+		fieldsStr, keyWord := filterParts[0], filterParts[1]
+		fields := strings.Split(fieldsStr, ".")
+
+		for i := qp.Start; i < end; i++ {
+			item := integrations[i]
+			bytes, err := json.Marshal(item)
+			if err != nil {
+				return nil, err
+			}
+			value, err := jsonparser.GetString(bytes, fields...)
```
cd1989: It's better to use `jsonpath` instead of `jsonparser` for this purpose. It's much simpler and cyclone already have this vendor.   Refer to https://github.com/caicloud/cyclone/blob/master/pkg/workflow/workload/pod/secret.go#L97

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -55,7 +57,35 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
...
+			if err != nil {
+				return nil, err
```
cd1989: When some broken resource object encountered, should not affect other normal ones. So here, I think should be `continue` and print a warning log.

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -38,11 +40,11 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
 	items := secrets.Items
 	integrations := []api.Integration{}
 	size := int64(len(items))
-	if pagination.Start >= size {
+	if qp.Start >= size {
 		return types.NewListResponse(int(size), integrations), nil
```
cd1989: Filter is not applied here.

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -55,7 +57,35 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
...
+		results = integrations[qp.Start:end]
+	} else {
+		filterParts := strings.Split(qp.Filter, "=")
+		if len(filterParts) != 2 {
+			return nil, cerr.ErrorUnknownInternal.Error("Filter query param %s is not correct", qp.Filter)
+		}
+		fieldsStr, keyWord := filterParts[0], filterParts[1]
+		fields := strings.Split(fieldsStr, ".")
+
+		for i := qp.Start; i < end; i++ {
```
cd1989: We should apply filters to all items, and then perform pagination for the remained items.

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -55,7 +57,35 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
...
+			item := integrations[i]
+			bytes, err := json.Marshal(item)
+			if err != nil {
+				return nil, err
+			}
+			value, err := jsonparser.GetString(bytes, fields...)
+			if err != nil {
+				return nil, err
+			}
+			if strings.Contains(strings.ToLower(value), strings.ToLower(keyWord)) {
```
cd1989: What if use give filter to numbers, datetime. By substring match, it would fail.

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -55,7 +57,35 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
 		integrations = append(integrations, *integration)
 	}
 
-	return types.NewListResponse(int(size), integrations[pagination.Start:end]), nil
+	results := []api.Integration{}
+	// Handle filter.
+	if qp.Filter == "" {
```
cd1989: For such case, return directly, would make the code clean.

if qp.Filter == "" {
  return ...
} 

...

pkg/server/handler/v1alpha1/integration.go
```diff
@@ -55,7 +57,35 @@ func ListIntegrations(ctx context.Context, tenant string, pagination *types.Pagi
 		integrations = append(integrations, *integration)
 	}
 
-	return types.NewListResponse(int(size), integrations[pagination.Start:end]), nil
+	results := []api.Integration{}
+	// Handle filter.
```
cd1989: I think such trivial comment can be removed.

pkg/server/handler/v1alpha1/admin.go
supereagle: Done

pkg/server/handler/v1alpha1/integration.go
supereagle: Done

supereagle:
Only support search and filter for needed APIs.  @cd1989  PTAL

PR #882, opened 2019-04-01
feat: skip ssl verify

moshuipan:

- skip ssl verify when use gitlab
- skip ssl verify when use git in worker

Fixes #

Reference to #

/cc @zhujian7 @wzqnls 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/scm/provider/github/github.go
+++ b/pkg/scm/provider/github/github.go
@@ -17,6 +17,8 @@ limitations under the License.
 package github
 
 import (
+	"context"
+	"crypto/tls"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


pkg/scm/provider/github/github.go
```diff
@@ -324,33 +326,86 @@ func (g *Github) DeleteWebHook(repoURL string, webHookUrl string) error {
...
+		c.BaseURL = baseEndpoint
+		c.UploadURL = uploadEndpoint
+	}
+	return c, nil
 }
 
 // NewTagFromLatest generate a new tag.
 func (g *Github) NewTagFromLatest(tagName, description, commitID, url string) error {
-	client := newClientByToken(g.scmCfg.Token)
+	client, _ := newClientByToken(g.scmCfg.Token, g.scmCfg.Server)
```
zhujian7: Is the second return value, maybe an `error`, need to check?

pkg/worker/scm/provider/git.go
```diff
@@ -121,6 +121,7 @@ func (g *Git) Clone(token, url, ref, destPath string) (string, error) {
 		}
 
 		cmds = []cmd{
+			// cmd{destPath, []string{"config", "--global", "http.sslVerify", "false"}},
```
zhujian7: Please delete it if not necessary

PR #889, opened 2019-04-03
feat: keep stage templates in cyclone system namespace

zhujian7:

Stage templates should be installed in the namespace which is consistent with Cyclone system namespace.

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/config/config.go
+++ b/pkg/server/config/config.go
@@ -43,7 +43,8 @@ type CycloneServerConfig struct {
 	// CreateBuiltinTemplates configures whether to create builtin stage templates while cyclone server start up.
 	CreateBuiltinTemplates bool `json:"create_builtin_templates"`
 
-	// SystemNamespace is the cyclone components installed namespace.
+	// SystemNamespace is the namespace where the Cyclone components installed in, and cyclone built-in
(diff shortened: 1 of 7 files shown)


pkg/server/config/config.go
```diff
@@ -118,4 +123,17 @@ func modifier(config *CycloneServerConfig) {
 		}
 	}
 
+	if config.SystemNamespace == "" {
```
cd1989: I'd prefer  to define a `Getter` to retrieve the namespace, and value set by `ENV` should have higher priority.

pkg/server/config/config.go
```diff
@@ -118,4 +123,17 @@ func modifier(config *CycloneServerConfig) {
...
+		log.Warningf("SystemNamespace not configured, try to retrieve value from ENV ")
+		config.SystemNamespace = os.Getenv(common.EnvSystemNamespace)
+	}
+}
+
+// validate the cyclone server config
+func validate(config *CycloneServerConfig) error {
+	if config.SystemNamespace == "" {
```
cd1989: Maybe we should fall back to default value `default` and print a warning,  instead of report error.

PR #890, opened 2019-04-03
feat: make creating default tenant configurable

zhujian7:

make creating system tenant configurable

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -87,9 +87,13 @@ func initialize(opts *Options) {
 	handler.Init(client)
 	log.Info("Init handlers succeed.")
 
-	err = v1alpha1.CreateAdminTenant()
-	if err != nil {
-		log.Fatalf("Create default tenant cyclone error %v", err)
+	if config.Config.InitDefaultTenant {
+		err = v1alpha1.CreateDefaultTenant()
+		if err != nil {
+			log.Fatalf("Create default cyclone tenant error %v", err)
+		}
+	} else {
+		log.Info("init_default_tenant is false, skip create default tenant")
 	}
 
 	if config.Config.CreateBuiltinTemplates {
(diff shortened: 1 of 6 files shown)


cmd/server/main.go
```diff
@@ -87,9 +87,13 @@ func initialize(opts *Options) {
 	handler.Init(client)
 	log.Info("Init handlers succeed.")
 
-	err = v1alpha1.CreateAdminTenant()
-	if err != nil {
-		log.Fatalf("Create default tenant cyclone error %v", err)
+	if config.Config.CreateSystemTenant {
```
cd1989: I'd prefer the term `default tenant` since it's also a normal tenant.

manifests/cyclone.yaml.template
```diff
@@ -80,6 +80,7 @@ data:
       },
       "create_builtin_templates": true,
       "system_namespace": "default",
+      "create_system_tenant": true,
```
cd1989: How about `init_default_tenant` ?

cmd/server/main.go
```diff
@@ -87,9 +87,13 @@ func initialize(opts *Options) {
...
+		err = v1alpha1.CreateAdminTenant()
```
cd1989: The same, should be `default tenant`

pkg/server/common/const.go
```diff
@@ -1,10 +1,10 @@
 package common
 
 const (
-	// AdminTenant is name of the system admin tenant, it's a default tenant created when Cyclone
+	// DefaultTenant is the name of the cyclone default tenant, it's a default tenant created when Cyclone
 	// start, and resources shared among all tenants would be placed in this tenant, such as stage
```
cd1989: > and resources shared among all tenants would be placed in this tenant, such as stage templates.

This should be removed now.

PR #893, opened 2019-04-04
feat(#892): search and filter stage templates

### LINKED ISSUE #892: Search and filter stage templates

/kind feature

**What happened**:

Stage templates can be searched by name and filtered by type.

/assign
/priority p0

supereagle:

Support to search and filter stage templates

Fixes #892

Reference to #

/cc @zhujian7  @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/template.go
+++ b/pkg/server/handler/v1alpha1/template.go
@@ -58,7 +58,7 @@ func ListTemplates(ctx context.Context, tenant string, includePublic bool, query
 	if query.Filter == "" {
 		results = items
 	} else {
-		// Support multiple filters rules: name and type, separated with comma.
+		// Support multiple filters rules: name or alias, and type, separated with comma.
 		filterParts := strings.Split(query.Filter, ",")
 		filters := make(map[string]string)
 		for _, part := range filterParts {
@@ -67,7 +67,7 @@ func ListTemplates(ctx context.Context, tenant string, includePublic bool, query
 				return nil, cerr.ErrorQueryParamNotCorrect.Error(query.Filter)
 			}
 
-			filters[kv[0]] = kv[1]
+			filters[kv[0]] = strings.ToLower(kv[1])
 		}
 
 		var selected bool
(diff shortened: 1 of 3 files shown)


pkg/server/handler/v1alpha1/template.go
```diff
@@ -45,7 +54,59 @@ func ListTemplates(ctx context.Context, tenant string, includePublic bool, query
...
+		// Support multiple filters rules: name or alias, and type, separated with comma.
+		filterParts := strings.Split(query.Filter, ",")
+		filters := make(map[string]string)
+		for _, part := range filterParts {
+			kv := strings.Split(part, "=")
+			if len(kv) != 2 {
+				return nil, cerr.ErrorQueryParamNotCorrect.Error(query.Filter)
+			}
+
+			filters[kv[0]] = strings.ToLower(kv[1])
```
cd1989: Why convert to lowercase, we can't filter case sensitively if so ?

pkg/server/handler/v1alpha1/template.go
```diff
@@ -45,7 +54,59 @@ func ListTemplates(ctx context.Context, tenant string, includePublic bool, query
...
+			selected = true
+			for key, value := range filters {
+				switch key {
+				case "name":
+					if !strings.Contains(item.Name, value) {
+						selected = false
+					}
+				case "alias":
+					if item.Annotations != nil {
+						if alias, ok := item.Annotations[meta.AnnotationAlias]; ok {
```
cd1989: If user specified `alias` filter, but a item doesn't have annotation alias, will it be accepted ? I think here if `item.Annotations` is nil or `ok` is false, also need set `selected` to false.

pkg/server/handler/v1alpha1/template.go
```diff
@@ -45,7 +54,59 @@ func ListTemplates(ctx context.Context, tenant string, includePublic bool, query
...
+							if !strings.Contains(alias, value) {
+								selected = false
+							}
+						}
+					}
+				case "type":
+					if item.Labels != nil {
```
cd1989: The same above. If no correct label found, should set `selected` to false.

pkg/server/handler/v1alpha1/template.go
```diff
@@ -45,7 +54,59 @@ func ListTemplates(ctx context.Context, tenant string, includePublic bool, query
 		items = append(items, publicTemplates.Items...)
 	}
 
-	size := int64(len(items))
+	results := []v1alpha1.Stage{}
+	if query.Filter == "" {
+		results = items
+	} else {
```
cd1989: Suggest to put `else` part to a new function.

pkg/server/handler/v1alpha1/template.go
supereagle: It is hard to user to make sure the case in filter, so we only provide case insensitive filter.

supereagle:
Have fixed comments, @cd1989  PTAL

PR #904, opened 2019-04-10
feat(workflow): add diagraph with workflow

huxiangtao:

import react digraph component for workflow 

image

Fixes #

Reference to #

/cc @qiuxiaolian 

**Release note**:

```release-note
NONE
```


web/src/components/workflow/component/graph.js
```diff
@@ -0,0 +1,291 @@
+import * as React from 'react';
```
qiuxiaolian: component 的文件命名首字母大写

PR #913, opened 2019-04-12
fix: apply correct resource requirements

cd1989:

Fix the problem that resource requirements set in wf spec is not used problem.

Fixes #

Reference to #

/cc @zhujian7 

**Release note**:

```release-note
NONE
```


supereagle:
@cd1989  Please fix ci failure.

PR #917, opened 2019-04-15
feat: support wfr cancel

cd1989:

Support WorkflowRun cancel

Fixes #

Reference to #

/cc @supereagle @zhujian7 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/controller/handlers/workflowrun/handler.go
+++ b/pkg/workflow/controller/handlers/workflowrun/handler.go
@@ -107,7 +107,7 @@ func (h *Handler) ObjectUpdated(obj interface{}) {
 	// the GC queue.
 	h.GCProcessor.Add(originWfr)
 
-	// If the WorkflowRun has already been terminated(Completed, Error, Cancel), send notifications if necessary,
+	// If the WorkflowRun has already been terminated(Completed, Failed, Cancelled), send notifications if necessary,
 	// otherwise directly skip it.
 	if workflowrun.IsWorkflowRunTerminated(originWfr) {
 		// Send notification after workflowrun terminated.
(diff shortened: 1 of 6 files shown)


pkg/workflow/workflowrun/operator.go
```diff
@@ -480,3 +482,31 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
...
+
+// Cancel cancels execution of the WorkflowRun, it will remove all pods of it.
+func (o *operator) Cancel() {
+	for stg, status := range o.wfr.Status.Stages {
+		if status.Pod == nil {
+			continue
+		}
+
+		// If the stage already in terminated status, skip it.
+		if status.Status.Phase == v1alpha1.StatusSucceeded ||
```
supereagle: This logic is frequently used, define a method for it?

pkg/workflow/workflowrun/operator.go
```diff
@@ -399,8 +410,9 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
 			if !wfrDeletion {
 				o.recorder.Eventf(o.wfr, corev1.EventTypeWarning, "GC", "Delete pod '%s' error: %v", status.Pod.Name, err)
 			}
+		} else {
+			log.WithField("ns", status.Pod.Namespace).WithField("pod", status.Pod.Name).Info("Pod deleted")
```
zhujian7: Why do this change?

pkg/workflow/controller/handlers/workflowrun/handler.go
```diff
@@ -107,7 +107,7 @@ func (h *Handler) ObjectUpdated(obj interface{}) {
 	// the GC queue.
 	h.GCProcessor.Add(originWfr)
 
-	// If the WorkflowRun has already been terminated(Completed, Error, Cancel), send notifications if necessary,
+	// If the WorkflowRun has already been terminated(Completed, Error), send notifications if necessary,
```
zhujian7: I think we also need to send notify even if it is canceled, @supereagle  WDYT?

pkg/workflow/workflowrun/utils.go
```diff
@@ -16,13 +16,13 @@ import (
 func resolveStatus(latest, update *v1alpha1.Status) *v1alpha1.Status {
 	// If the latest status is already a terminated status (Completed, Error), no need to update
```
zhujian7: terminated status should be `Succeeded, Failed and Cancelled`

pkg/workflow/controller/handlers/workflowrun/handler.go
supereagle: In my next PR, will send notification for all wfrs.

PR #937, opened 2019-04-26
fix: support different token format in git resolver

cd1989:

- Support both <user>:<password> and <token> in git resolver
- URL encode token 

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/build/resolver/git/entrypoint.sh
+++ b/build/resolver/git/entrypoint.sh
@@ -9,7 +9,7 @@ USAGE=$(cat <<-END
         $ docker run -it --rm \\
             -e GIT_URL=https://github.com/caicloud/cyclone.git \\
             -e GIT_REVISION=master \\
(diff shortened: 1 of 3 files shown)


build/resolver/git/entrypoint.sh
```diff
@@ -28,7 +28,9 @@ USAGE=$(cat <<-END
        'refs/pull/1/merge', but for Gitlab, composite revision is necessary, such as
        'refs/merge-requests/1/head:master'.
      - GIT_TOKEN [Optional] For public repository, no need provide this token, but for
-       private repository, this should be provided.
+       private repository, this should be provided. Token here support 2 different formats:
```
supereagle: Prefer to change the argument to `GIT_AUTH` to for these 2 formats.

PR #948, opened 2019-04-29
fix: support empty argument value and escape line break

cd1989:

- Support empty argument value "" in engine
- Escape '\n' in argument

Fixes #

Reference to #

/cc @supereagle 

**Release note**:

```release-note
NONE
```


pkg/workflow/workload/pod/builder.go
```diff
@@ -116,6 +116,8 @@ func (m *Builder) ResolveArguments() error {
 	// Escape double quotes
```
supereagle: // Escape double quotes and new lines.

pkg/workflow/workload/pod/builder.go
```diff
@@ -116,6 +116,8 @@ func (m *Builder) ResolveArguments() error {
...
 	for k, v := range parameters {
 		parameters[k] = strings.Replace(v, "\"", "\\\"", -1)
+		parameters[k] = strings.Replace(parameters[k], "\n", "\\n", -1)
+		log.Errorf("Key: %s, Value: %s", k, parameters[k])
```
supereagle: log error?

cd1989: Debug code, I will remove it.

PR #956, opened 2019-05-06
improve: make open control cluster configurable

cd1989:

Make it configurable to open control cluster for workflow execution. So that, we can deploy cyclone with everything ready for workflow execution.

Fixes #

Reference to #

/cc @zhujian7

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/integration.go
+++ b/pkg/server/handler/v1alpha1/integration.go
@@ -163,41 +163,47 @@ func OpenClusterForTenant(in *api.Integration, tenantName string) (err error) {
 		cluster.Namespace = svrcommon.TenantNamespace(tenant.Name)
 		err = svrcommon.CreateNamespace(tenant.Name, client)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/server/handler/v1alpha1/integration.go
```diff
@@ -163,41 +163,47 @@ func OpenClusterForTenant(in *api.Integration, tenantName string) (err error) {
...
-			log.Errorf("get pvc %s error %v", cluster.PVC, err)
+			log.Errorf("Get pvc %s error %v", cluster.PVC, err)
 			return
 		}
 	} else {
-		// create pvc
 		if tenant.Spec.PersistentVolumeClaim.Size == "" {
 			tenant.Spec.PersistentVolumeClaim.Size = config.Config.DefaultPVCConfig.Size
 		}
+		if tenant.Spec.PersistentVolumeClaim.StorageClass == "" {
```
zhujian7: This needs more discussion If we don't specify a StorageClass, It will use the default in the cluster.
So I don't know if the change is needed.

PR #986, opened 2019-05-14
feat: support multiple workflow controller

cd1989:

Add controller instance id to workflow controller. If instance set, controllers for `pod`, `workflowrun`, `workflowruntrigger` will only watch resources with the correct instance label. And pod, wfr created by the workflow controller will also have controller instance label set.

To keep back-compatibility, if no instance id set, no label would be applied, and controllers would watch resources without the controller instance label.

Note: only workflow controller supports multiple instance. Cyclone sever would ignore the controller instance label.

To set instance id, set environment variable `CONTROLLER_INSTANCE_NAME`, for example: `cyclone`

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/common/const.go
+++ b/pkg/common/const.go
@@ -13,4 +13,7 @@ const (
 
 	// ControlClusterName is name of the master cluster
 	ControlClusterName = "cyclone-control-cluster"
+
+	// ControllerInstanceEnvName is environment name for workflow controller instance
+	ControllerInstanceEnvName = "CONTROLLER_INSTANCE_NAME"
 )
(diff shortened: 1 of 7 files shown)


supereagle:
Do we want to support multiple workflow controller in this way? I think this is not a good practice.

cd1989:
> Do we want to support multiple workflow controller in this way? I think this is not a good practice. Is there any requirement for this feature?

One scene is: test workflow controller locally, when I start a controller locally, it will conflict with the controller in the environment or in other developer's local ones. 

And argo also provide such feature.

zhujian7:
> One scene is: test workflow controller locally, when I start a controller locally, it will conflict with the controller in the environment or in other developer's local ones.

That sounds awesome!

PR #1021, opened 2019-05-26
fix: list workflow triggers belonged to the specific workflow

zhujian7:

The list wfts API listed all workflow triggers and did not filter triggers that do not belong to the specific workflow. fix that.

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/helper.go
+++ b/pkg/server/handler/v1alpha1/helper.go
@@ -341,19 +341,24 @@ func InjectWorkflowOwnerRefModifier(tenant, project, wf string, object interface
 	return nil
 }
 
-// WorkflowRunModifier is a modifier of create cyclone workflowrun resources.
-// It will add workflow labels for workflowrun.
-func WorkflowRunModifier(tenant, project, wf string, object interface{}) error {
-	wfr, ok := object.(*v1alpha1.WorkflowRun)
-	if !ok {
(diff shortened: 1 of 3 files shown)


pkg/server/handler/v1alpha1/workflowtrigger.go
```diff
@@ -56,8 +56,15 @@ func ListWorkflowTriggers(ctx context.Context, tenant, project, workflow string,
 	}
 
 	items := workflowTriggers.Items
+	var result []v1alpha1.WorkflowTrigger
```
supereagle: I'd prefer to add workflow label, then we can directly filter them by label selector.

zhujian7: ok

zhujian7:
@supereagle PTAL

pkg/server/handler/v1alpha1/helper.go
```diff
@@ -341,19 +341,24 @@ func InjectWorkflowOwnerRefModifier(tenant, project, wf string, object interface
 	return nil
 }
 
-// WorkflowRunModifier is a modifier of create cyclone workflowrun resources.
-// It will add workflow labels for workflowrun.
-func WorkflowRunModifier(tenant, project, wf string, object interface{}) error {
```
supereagle: `WorkflowRunModifier()` is still in use, why remove it?

PR #1041, opened 2019-05-28
fix: templates web

cd1989:

- template item alignment
- artifacts display name

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```


web/src/locale/zh-CN.yaml
```diff
@@ -170,6 +170,8 @@ stage:
   input:
     arguments: 参数
   outputs: 输出
+  output:
+    artifacts: Artifacts
```
supereagle: 一般的叫法是`产出物`?

PR #1048, opened 2019-05-29
fix(stageTemplate): fix configuration cmd value & resource options

huxiangtao:

fix configuration cmd value 
set resources options optional access props
replace cmd regex
edit css of template list item action button

Fixes #

Reference to #

/cc @cd1989 

**Release note**:

```release-note
NONE
```


web/src/components/stageTemplate/Detail/Inputs.js
```diff
@@ -7,7 +7,7 @@ const valueRender = (v, item) => {
   if (item.name === 'cmd') {
     return (
       <SyntaxHighlighter language="bash" style={duotoneLight}>
-        {v.replace(/;\s+/g, ';\n')}
+        {v && v.replace(/;\s*/g, ';\n')}
```
cd1989: If v is empty, do not use `SyntaxHighlighter`

PR #1053, opened 2019-05-30
fix(workflow): modify stage template select options key & alias

huxiangtao:

replace stage template select options key & alias with unique key which is composed by <namespace>-<templatename>
image

Fixes #

Reference to #

/cc @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/web/src/components/workflow/component/stage/AddStage.js
+++ b/web/src/components/workflow/component/stage/AddStage.js
@@ -110,21 +110,15 @@ class AddStage extends React.Component {
       >
         {templates.map(o => {
           const name = _.get(o, 'metadata.name');
-          const alias = _.get(o, [
-            'metadata',
-            'annotations',
-            'cyclone.dev/alias',
-          ]);
           const namespace = _.get(o, 'metadata.namespace', '');
           const key = namespace ? `${namespace}-${name}` : name;
-          const templateAlias = namespace ? `${namespace}-${alias}` : alias;
           return (
             <Option value={name} key={key}>
               {_.get(o, ['metadata', 'labels', 'cyclone.dev/builtin']) ===
               'true'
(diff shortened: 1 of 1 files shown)


web/src/components/workflow/component/stage/AddStage.js
```diff
@@ -110,12 +110,21 @@ class AddStage extends React.Component {
...
           return (
-            <Option value={name} key={name}>
+            <Option value={name} key={key}>
               {_.get(o, ['metadata', 'labels', 'cyclone.dev/builtin']) ===
               'true'
-                ? intl.get(`template.kinds.${name.replace('-template', '')}`)
-                : _.get(o, ['metadata', 'annotations', 'cyclone.dev/alias'])}
+                ? intl.get(`template.kinds.${name.replace('-template', '')}`) ||
+                  key
+                : templateAlias}
```
cd1989: No need to use namespace prefix for displaying.  The list contains only templates belonging to the tenant (same namepace) together with system templates.

web/src/components/workflow/component/stage/AddStage.js
```diff
@@ -110,12 +110,21 @@ class AddStage extends React.Component {
...
         {templates.map(o => {
           const name = _.get(o, 'metadata.name');
+          const alias = _.get(o, [
+            'metadata',
+            'annotations',
+            'cyclone.dev/alias',
+          ]);
+          const namespace = _.get(o, 'metadata.namespace', '');
+          const key = namespace ? `${namespace}-${name}` : name;
+          const templateAlias = namespace ? `${namespace}-${alias}` : alias;
```
cd1989: We don't alias in Cyclone. Use name directly.

PR #1057, opened 2019-05-30
fix(graph): fix intl problem

huxiangtao:

Add your description

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```

### DIFF
--- a/web/src/locale/en-US.yaml
+++ b/web/src/locale/en-US.yaml
@@ -48,6 +48,7 @@ sideNav:
   workflow: Workflow
   integration: Integration Center
 workflow:
+  addStage: Add Stage 
   recentVersion: Recent version
   creator: Creator
   creationTime: Creation time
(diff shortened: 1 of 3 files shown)


web/src/locale/en-US.yaml
```diff
@@ -48,6 +48,7 @@ sideNav:
   workflow: Workflow
   integration: Integration Center
 workflow:
+  addStage: add stage
```
cd1989: Add Stage

web/src/locale/zh-CN.yaml
```diff
@@ -48,6 +48,7 @@ sideNav:
   workflow: 流水线
   integration: 集成中心
 workflow:
+  addStage: 添加 stage
```
cd1989: 添加阶段

PR #1070, opened 2019-05-31
feat: support arbitrary resource type

cd1989:

- Support arbitrary resource type registered

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/cyclone/v1alpha1/resource.go
+++ b/pkg/apis/cyclone/v1alpha1/resource.go
@@ -2,6 +2,13 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+const (
+	// ImageResourceType represents image in docker registry
+	ImageResourceType = "Image"
+	// GitResourceType represents git repo in SCM
+	GitResourceType = "Git"
+)
+
 // +genclient
 // +genclient:noStatus
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -15,20 +22,6 @@ type Resource struct {
 	// Resource specification
 	Spec ResourceSpec `json:"spec"`
 }
-
-// ResourceType defines resource type
-type ResourceType string
-
-const (
-	// ImageResourceType represents image in docker registry
-	ImageResourceType = "Image"
-	// GitResourceType represents git repo in SCM
-	GitResourceType = "Git"
-	// SvnResourceType represents svn repo in SCM
-	SvnResourceType = "Svn"
-	// GeneralResourceType represents general resource
-	GeneralResourceType = "General"
-)
 
 // ResourcePullPolicy indicates resource pull policy
 type ResourcePullPolicy string
@@ -45,7 +38,7 @@ type ResourceSpec struct {
 	// Image to resolve this kind of resource.
 	Resolver string `json:"resolver,omitempty"`
 	// Resource type, e.g. image, git, kv, general.
-	Type ResourceType `json:"type"`
+	Type string `json:"type"`
 	// Persistent resource to PVC.
 	Persistent *Persistent `json:"persistent"`
 	// Parameters of the resource
--- a/pkg/server/handler/v1alpha1/resourcetypes.go
+++ b/pkg/server/handler/v1alpha1/resourcetypes.go
@@ -11,40 +11,12 @@ import (
 
 	"github.com/caicloud/cyclone/pkg/apis/cyclone/v1alpha1"
 	"github.com/caicloud/cyclone/pkg/common"
-	"github.com/caicloud/cyclone/pkg/meta"
 	svrcommon "github.com/caicloud/cyclone/pkg/server/common"
 	"github.com/caicloud/cyclone/pkg/server/handler"
 	"github.com/caicloud/cyclone/pkg/server/types"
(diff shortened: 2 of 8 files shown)


cd1989:
retest

pkg/server/handler/v1alpha1/resourcetypes.go
```diff
@@ -11,49 +11,21 @@ import (
...
-
 // ListResourceTypes ...
 func ListResourceTypes(ctx context.Context, tenant string, operation string) (*types.ListResponse, error) {
 	namespaces := []string{svrcommon.TenantNamespace(tenant)}
 	if tenant != svrcommon.DefaultTenant {
 		namespaces = append(namespaces, common.GetSystemNamespace())
 	}
 
-	resources, err := getResourceTypes(namespaces, operation)
+	resources, err := wfcommon.GetResourceTypes(handler.K8sClient, namespaces, operation)
```
supereagle: It seems that there is only one `handler.K8sClient`, why not put this into `GetResourceTypes()`, then there is no need to pass anywhere.

At the same time, I think `GetResourceTypes()` is better to kept in this file, rather than moved into workflow common package.

pkg/workflow/controller/config.go
```diff
@@ -21,12 +19,6 @@ const (
 	// ConfigFileKey is key of config file in ConfigMap
 	ConfigFileKey = "workflow-controller.json"
 
-	// GitResolverImage is key of git source resolver image in config file
```
supereagle: Also need to remove them from configmap in helm chart and release yaml.

pkg/workflow/workload/pod/builder.go
```diff
@@ -344,9 +335,16 @@ func (m *Builder) ResolveInputResources() error {
 			})
 		}
 
+		// Get resource resolver for the given resource type. If the resource has resolver set, use it directly,
+		// otherwise get resolver from registered resource types.
+		resolver, err := common.GetResourceResolver(m.client, resource)
```
supereagle: If the client is for user clusters, then it can not get resource types.

pkg/apis/cyclone/v1alpha1/resource.go
```diff
@@ -16,20 +23,6 @@ type Resource struct {
 	Spec ResourceSpec `json:"spec"`
 }
 
-// ResourceType defines resource type
-type ResourceType string
```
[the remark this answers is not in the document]
cd1989: No need to define type, since name is not static identify for resource type. Resource type is defined in Resource Type object.

pkg/server/handler/v1alpha1/resourcetypes.go
cd1989: `common` provides utility, shouldn't depends on global variables like `handler.K8sClient`, otherwise it can't be used in other places if `handler.K8sClient` not initialized.

> I think GetResourceTypes() is better to kept in this file, rather than moved into workflow common package.

It will also be used in workflow engine, workflow engine shouldn't call top layer functions.

pkg/workflow/workload/pod/builder.go
cd1989: This client is used to be control cluster client.

pkg/workflow/controller/config.go
cd1989: helm already updated, for release, need more considerations, i'd like to handle it in another PR, cyclone is cyclone, cps is cps

pkg/apis/cyclone/v1alpha1/resource.go
```diff
@@ -2,6 +2,13 @@ package v1alpha1
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+const (
+	// ImageResourceType represents image in docker registry
+	ImageResourceType = "Image"
+	// GitResourceType represents git repo in SCM
+	GitResourceType = "Git"
```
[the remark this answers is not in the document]
cd1989: Not used, so no need to define.

pkg/server/handler/v1alpha1/resourcetypes.go
supereagle: > `common` provides utility, shouldn't depends on global variables like `handler.K8sClient`, otherwise it can't be used in other places if `handler.K8sClient` not initialized.

Make sense.

> It will also be used in workflow engine, workflow engine shouldn't call top layer functions.

It is used by server and workflow engine, Why not put inot `pkg/common`?

pkg/apis/cyclone/v1alpha1/resource.go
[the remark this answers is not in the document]
supereagle: It seems that `ImageResourceType` and `GitResourceType` are only used in `pkg/workflow/workload/pod/builder_test.go`. If so, I think there is no need to define const here.

pkg/server/handler/v1alpha1/resourcetypes.go
cd1989: Resource Type is concept of Workflow Engine, it's natural to be defined in Workflow Engine layer. And Cyclone Server already use lots pkg from Workflow Engine. I'd like to make top level 'pkg/common' lightweight

supereagle: ok, I think we need to uniformly handle these common packages later.

PR #1073, opened 2019-06-01
feat: bind resource type to integration type

cd1989:

- Bind resource type to integration type, so that we can know how to use integration for resource.

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```


web/src/locale/zh-CN.yaml
```diff
@@ -213,6 +213,7 @@ resource:
   type: 资源类型
   resolver: 资源解析器
   operations: 支持操作
+  bingding: 集成服务绑定
```
supereagle: typo?

PR #1121, opened 2019-06-06
 fix(stage): stage config command

qiuxiaolian:

- modify command

Fixes #

Reference to #

/cc @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/web/src/components/workflow/component/stage/StageField.js
+++ b/web/src/components/workflow/component/stage/StageField.js
@@ -79,14 +79,6 @@ class StageField extends React.Component {
                         validate={required}
                         formItemLayout={drawerFormItemLayout}
                       />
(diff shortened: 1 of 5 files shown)


web/src/components/workflow/component/stage/StageField.js
```diff
@@ -82,14 +82,14 @@ class StageField extends React.Component {
                       {/* // TODO: 暂时不展示此字段 */}
                       {/* <Field
                         label={'ENTRYPOINT'}
-                        name={`${specKey}.spec.containers.${index}.command`}
+                        name={`${specKey}.spec.containers.${index}.args`}
```
cd1989: Here it should be `command` for `ENTRYPOINT`. Please remove the this part as not used.

PR #1145, opened 2019-06-18
fix: generate image random tag

zhujian7:

Generate image random tag for scm triggered wfr

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/webhook.go
+++ b/pkg/server/handler/v1alpha1/webhook.go
@@ -156,10 +156,13 @@ func createWorkflowRun(tenant string, wft v1alpha1.WorkflowTrigger, data *scm.Ev
 
 	log.Infof("Trigger wft %s with event data: %v", wft.Name, data)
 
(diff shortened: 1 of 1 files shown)


pkg/server/handler/v1alpha1/webhook.go
```diff
@@ -184,10 +185,14 @@ func createWorkflowRun(tenant string, wft v1alpha1.WorkflowTrigger, data *scm.Ev
 		return err
 	}
 
+	if tag == "" {
```
supereagle: Should add this logic in https://github.com/caicloud/cyclone/pull/1145/files#diff-4db6e7deb63057dec099088e5e228cfbR160.

```
	suffix := rand.String(5)
	name := fmt.Sprintf("%s-%s", wfName, suffix)
	alias := name
	if tag != "" {
		alias = tag
	} else {
		tag = suffix
	}
```

zhujian7:
@supereagle PTAL

PR #1199, opened 2019-07-23
feat(server): support dry-run for creating integration

zhujian7:

- support dry-run for creating integration, we can use this feature to validate integration before the creation
- refactor gitlab error conversion

Fixes #

Reference to #

/cc @cd1989 @supereagle 

**Release note**:

```release-note
NONE
```


cd1989:
The validation APIs only works for some specific integration types, but always return true for other types. It's not a good design, I don't think we should create it. We can validate in the create API and return proper errors when validation failed.

PR #1218, opened 2019-08-05
fix: validate bitbucket token

zhujian7:

Fix bitbucket token validation will always return true.

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```


pkg/server/biz/scm/bitbucket/server/server.go
```diff
@@ -75,9 +75,13 @@ func (b *BitbucketServer) GetToken() (string, error) {
 
 // CheckToken checks whether the token has the authority of repo by trying ListRepos with the token.
 func (b *BitbucketServer) CheckToken() error {
-	if _, err := b.listReposInner(false); err != nil {
+	repos, err := b.listReposInner(false)
+	if err != nil {
 		return err
 	}
+	if len(repos) == 0 {
```
supereagle: Do you consider that there is no repos?

PR #1247, opened 2019-08-12
fix: send notification only once

zhujian7:

Fix workflow controller sent notification multiple times.

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```


pkg/workflow/controller/handlers/workflowrun/handler.go
```diff
@@ -172,7 +185,15 @@ func (h *Handler) ObjectDeleted(obj interface{}) {
 // * notification endpoint is configured
 // * without notification sent label
 func (h *Handler) sendNotification(wfr *v1alpha1.WorkflowRun) error {
-	if meta.LabelExists(wfr.Labels, meta.LabelWorkflowRunNotificationSent) {
+	h.notifyLock.Lock(wfr.Name)
+	defer h.notifyLock.Unlock(wfr.Name)
```
cd1989: Seems no need to use lock, controller handles event in a single thread one by one.

zhujian7: Have fixed, PTAL @cd1989

PR #1249, opened 2019-08-13
fix: git clone from bitbucket failed

zhujian7:

Git clone with "--depth" option will fail when the server is Bitbucket which version less than v0.6.4(This version is not guaranteed to be accurate, I tested v0.6.4 support "--depth", but v0.5.4.9 not support), so we will not use "--depth" when the server is Bitbucket.

Fixes #

Reference to #

/cc @supereagle @cd1989 

**Release note**:

```release-note
NONE
```


build/resolver/git/entrypoint.sh
```diff
@@ -180,7 +187,7 @@ pull() {
         }
 
         echo "Fetch $SCM_REVISION from origin"
-        git fetch -v --depth=1 origin $SCM_REVISION
+        git fetch -v ${GIT_DEPTH_OPTION} origin $SCM_REVISION
```
cd1989: Use `${GIT_DEPTH_OPTION:-}` to use default empty value if not variable set, otherwise

zhujian7: PTAL

PR #1321, opened 2019-11-08
fix: deleting pod stuck on Terminating status

zhujian7:
WorkflowRun GC SHOULD Wait all workflowRun related workload pods
deleting completed, then start gc pod to clean date on PV.Otherwise,
if the path which is used by workload pods in the PV is deleted
before workload pods deletion, the pod deletion process will get
stuck on Terminating status.

Add your description

Fixes #

Reference to #

/cc @supereagle @pendoragon @orainxiong 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/workflowrun/operator.go
+++ b/pkg/workflow/workflowrun/operator.go
@@ -422,6 +422,7 @@ func (o *operator) Reconcile() error {
 // GC would performed silently, without event recording and status updating.
 func (o *operator) GC(lastTry, wfrDeletion bool) error {
 	wg := sync.WaitGroup{}
+	allPodsFinished := true
 	// For each pod created, delete it.
 	for stg, status := range o.wfr.Status.Stages {
 		// For non-terminated stage, update status to cancelled.
@@ -468,6 +469,7 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
 				for {
 					select {
 					case <-timeout:
+						allPodsFinished = false
 						log.WithField("ns", namespace).WithField("pod", podName).Warn("Pod deletion timeout")
 						return
 					case <-ticker.C:
@@ -486,6 +488,15 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
 	// Otherwise, if the path which is used by workload pods in the PV is deleted before workload pods deletion,
 	// the pod deletion process will get stuck on Terminating status.
 	wg.Wait()
+
+	// If there are pods not finished and this is not the last gc try process, we will not start gc pod to clean
+	// data on PV. The last gc try process will ensure data could be cleaned.
+	if !allPodsFinished && !lastTry {
+		if !wfrDeletion {
+			o.recorder.Eventf(o.wfr, corev1.EventTypeWarning, "GC", "There are stage pods not Finished")
+		}
+		return nil
+	}
 
 	// Get execution context of the WorkflowRun, namespace and PVC are defined in the context.
 	executionContext := GetExecutionContext(o.wfr)
--- a/pkg/workflow/workflowrun/operator.go
+++ b/pkg/workflow/workflowrun/operator.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
@@ -420,6 +421,7 @@ func (o *operator) Reconcile() error {
 // - 'wfrDeletion' indicates whether the GC is performed because of WorkflowRun deleted. In this case,
 // GC would performed silently, without event recording and status updating.
 func (o *operator) GC(lastTry, wfrDeletion bool) error {
+	wg := sync.WaitGroup{}
 	// For each pod created, delete it.
 	for stg, status := range o.wfr.Status.Stages {
 		// For non-terminated stage, update status to cancelled.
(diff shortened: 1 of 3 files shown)


pkg/workflow/workflowrun/operator.go
```diff
@@ -454,10 +456,37 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
...
+							log.WithField("ns", namespace).WithField("pod", podName).Info("Pod deleted")
+							return
+						}
+					}
+				}
+			}(status.Pod.Namespace, status.Pod.Name)
 		}
 	}
 
+	// Wait all workflowRun related workload pods deleting completed, then start gc pod to clean date on PV.
```
orainxiong: *clean date* ---> *clean data*

pkg/workflow/workflowrun/operator.go
```diff
@@ -454,10 +456,37 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
...
+					select {
+					case <-timeout:
+						log.WithField("ns", namespace).WithField("pod", podName).Warn("Pod deletion timeout")
+						return
+					case <-ticker.C:
+						_, err := o.clusterClient.CoreV1().Pods(namespace).Get(podName, metav1.GetOptions{})
+						if err != nil && errors.IsNotFound(err) {
+							log.WithField("ns", namespace).WithField("pod", podName).Info("Pod deleted")
+							return
+						}
```
supereagle: Miss the logic of https://github.com/caicloud/cyclone/pull/1321/files#diff-9bb1eecfaf9ae747dcffe86457ab7f58R450-R457

Any way to merge the first deletion and the retries.

pkg/workflow/workflowrun/operator.go
```diff
@@ -454,10 +456,37 @@ func (o *operator) GC(lastTry, wfrDeletion bool) error {
...
+					}
+				}
+			}(status.Pod.Namespace, status.Pod.Name)
 		}
 	}
 
+	// Wait all workflowRun related workload pods deleting completed, then start gc pod to clean date on PV.
+	// Otherwise, if the path which is used by workload pods in the PV is deleted before workload pods deletion,
+	// the pod deletion process will get stuck on Terminating status.
+	wg.Wait()
```
pendoragon: does this mean that if a workflowrun pod stuck on terminating for some reason, gc would pause for 5 min?

zhujian7: Yes，I don't know how to judge the error more precisely.

pkg/workflow/workflowrun/operator.go
zhujian7: ok

pkg/workflow/workflowrun/operator.go
zhujian7: I think it's not necessary, since if deleting pod gets result `error not found`, the for loop will continue.

pkg/workflow/workflowrun/operator.go
pendoragon: for every finished wfr pod, a new gc pod would run to clean its subpath?

zhujian7: One wfr may start multiple pods(one stage a pod), after all pods started by the wfr finished, a new gc pod would run to clean data.

pendoragon: so if any of the pods started by a wfr stuck on terminating for more than 5 min, the gc for this wfr will fail and the subpath will remain in the pv indefinitely(since we don't have any way of triggering gc for this particular wfr again)?

zhujian7: No, 5 min is a timeout threshold, if timeout, cyclone will start a  gc pod to delete subpath.

pendoragon: got it. so this does not solve the issue once and for all. it fixes the issue for most of the scenarios while we may still expect glitches in some corner cases?

zhujian7: Right, Do you think it's acceptable?

pendoragon: I think this might be the best we can get with minimal effort. And in cases where wfr pods stuck on terminating for other reasons, chances are we have to handle it manually(and it is stuck anyways).

Just off the top of my head, is it possible if we have the gc run periodically and check all of the subpath in the pvc and delete those which does not have a corresponding pod(suppose we have a one to one mapping between pod name and subpath name) so that we will never delete any subpath before pod cleanup? Not sure if this way we can solve the issue. Anyways, this is not a trivial change that is appropriate for a patch version. 

So I think this is good enough for now, but hopefully we can come up with a better solution that will eliminate the issue once and for all.

zhujian7: SGTM, and I submit [another commit](https://github.com/caicloud/cyclone/pull/1321/commits/fb2cea27f8cc2ac06e75dac0c13957c6384903e1) to ensure that **only the last** retry GC process (cyclone GC, deleting stage pods and deleting data in PV, has a retry mechanism) will delete data in PV even if there are stage pods not finished, otherwise, cyclone will wait for the next retry GC to execute. 

Although this commit can not eliminate the issue once and for all, It will reduce the probability of the corner case occurrences.

zhujian7:
@supereagle PTAL

pkg/workflow/workflowrun/operator.go
zhujian7: @pendoragon PTAL

pendoragon: LGTM

PR #1350, opened 2019-12-23
fix(*): fix workflow required field and undefined func

qiuxiaolian:

- Whether the field is required
- Clickable secondary breadcrumbs
- log  component bug fix 

Fixes #

Reference to #

/cc @ladjzero

**Release note**:

```release-note
NONE
```

### DIFF
--- a/web/src/components/workflow/component/stage/TemplateStage.js
+++ b/web/src/components/workflow/component/stage/TemplateStage.js
@@ -98,9 +98,7 @@ class TemplateStage extends React.Component {
                           }
                           tooltip={_.get(argDes, r.name)}
                           hasFeedback
-                          required={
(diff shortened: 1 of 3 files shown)


web/src/components/workflow/component/stage/TemplateStage.js
```diff
@@ -98,7 +98,9 @@ class TemplateStage extends React.Component {
                           }
                           tooltip={_.get(argDes, r.name)}
                           hasFeedback
-                          required
+                          required={
+                            r.name === 'output-artifact-path' ? false : true
```
ladjzero:
```suggestion
                            r.name !== 'output-artifact-path'
```

PR #1353, opened 2019-12-25
chore: get tenant info from query if header omitted

zhujian7:

Add your description

Fixes #

Reference to #

/cc @supereagle @qiuxiaolian 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/workflowrun.go
+++ b/pkg/server/handler/v1alpha1/workflowrun.go
@@ -353,7 +353,14 @@ func receiveContainerLogStream(tenant, project, workflow, workflowrun, stage, co
 }
 
 // GetContainerLogStream gets real-time logs of a certain stage.
-func GetContainerLogStream(ctx context.Context, project, workflow, workflowrun, tenant, stage string) error {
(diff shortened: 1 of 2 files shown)


pkg/server/handler/v1alpha1/workflowrun.go
```diff
@@ -353,7 +353,14 @@ func receiveContainerLogStream(tenant, project, workflow, workflowrun, stage, co
 }
 
 // GetContainerLogStream gets real-time logs of a certain stage.
-func GetContainerLogStream(ctx context.Context, project, workflow, workflowrun, tenant, stage string) error {
+func GetContainerLogStream(ctx context.Context, project, workflow, workflowrun, tenant, stage, qt string) error {
+	// If there is no tenant info in Header, we will get it from query, since front end is hard to
```
supereagle: This is a little tricky, we can use HTTP request and upgrade to websocket. @qiuxiaolian

qiuxiaolian: This problem can be solved by setting a proxy. However, the version of the project's create-react-app is too low, and modifying the configuration is cumbersome and requires a relatively large price. Can be resolved after subsequent upgrades

PR #1460, opened 2020-06-30
[EIP]:fix(cyclone):judgement use eip or vip according to user cluster

quasimodo7614:

fix(cyclone):judgement use eip or vip according user cluster

Fixes #

Reference to #

/cc @zhujian7 @tosone 

**Release note**:

```release-note
NONE
```


pkg/workflow/common/cluster.go
```diff
@@ -17,3 +17,15 @@ func GetExecutionClusterClient(wfr *v1alpha1.WorkflowRun) kubernetes.Interface {
 
 	return store.GetClusterClient(cluster)
 }
+
+// IsMaster judge weather this wkf in master cluster or not
```
zhujian7, line 21: `IsMaster judge whether this wfr will be executed in the master cluster or not` is better?

pkg/workflow/controller/config.go
```diff
@@ -47,6 +47,9 @@ type WorkflowControllerConfig struct {
 	ExecutionContext ExecutionContext `json:"execution_context"`
 	// CycloneServerAddr is address of the Cyclone Server
 	CycloneServerAddr string `json:"cyclone_server_addr"`
+	// CycloneServerAddrEIP is the address of the Cyclone Server . when not in master cluster,
+	// and the net solution is EIP , use this Addr  rather than CycloneServerAddr
```
zhujian7, line 51: delete the unnecessary "space" between `use this Addr` and `rather than ...`?

zhujian7:
I don't recommend to add a new `master` concept in the repo. Can we keep using `ControlCluster`?

PR #1467, opened 2020-07-13
fix: status wrong when task processed in the delegation api

zhujian7:

If the delegation task processed in the delegation API and update the task status to Succeeded/Failed status, we should not update it to Waiting(since the task has already been processed), otherwise, the task will be stuck.

Fixes #

Reference to #

/cc @hyy0322 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/workflowrun/workload.go
+++ b/pkg/workflow/workflowrun/workload.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/caicloud/cyclone/pkg/apis/cyclone/v1alpha1"
 	"github.com/caicloud/cyclone/pkg/k8s/clientset"
+	"github.com/caicloud/cyclone/pkg/util"
 	"github.com/caicloud/cyclone/pkg/util/retry"
 	"github.com/caicloud/cyclone/pkg/workflow/workload/delegation"
 	"github.com/caicloud/cyclone/pkg/workflow/workload/pod"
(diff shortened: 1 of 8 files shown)


hyy0322:
Delegation task is something like calling external service like jenkins to do some work?

hyy0322:
Why deletion task processed twice?

pkg/workflow/workflowrun/workload.go
```diff
@@ -144,6 +144,22 @@ func (p *WorkloadProcessor) processDelegation() error {
...
+
+	// If the task has already been processed by the above RESTful API task, we should not update the status of the task to Waiting
+	latestWfr, err := p.client.CycloneV1alpha1().WorkflowRuns(p.wfr.Namespace).Get(p.wfr.Name, metav1.GetOptions{})
+	if err != nil {
+		log.WithField("wfr", p.wfr.Name).Error("Get latest wfr failed")
+		return err
+	}
+	// The task maybe has been processed by the delegation task and status not needed to be set to Waiting
+	if stg, ok := latestWfr.Status.Stages[p.stg.Name]; ok {
+		if stg.Status.Phase == v1alpha1.StatusSucceeded ||
```
supereagle, line 156: I remember that we have defined a method for this status judgement. If not, I think we should define one for this.

supereagle:
ping @zhujian7

pkg/workflow/workflowrun/workload.go
zhujian7: OK, I will refactor it.

zhujian7:
> ping @zhujian7

@supereagle  PTAL

PR #1468, opened 2020-07-14
fix(controller): pod stuck on terminating while related wf deleted

zhujian7:

When we delete a workflow that has pods running, the pods will remain in Terminating status after the workflow deletion.

Old logic details: 
1. **Delete the workflow**, will delete the related workflowRuns(OwnerReferences)
2. Delete the workflowRun will delete the related pods, **and wait until the pods deleting completed**
3. Delete the pod **will get the workflow** and update status of the workflowRun, workflow has already been deleted so failed.

Actually, when the workflow is under deleting(deletionTimestamp is not nil)，we do not need to update status of the workflowRun, so the problem will be solved.

Fixes #

Reference to #

/cc @your-reviewer

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/controller/controllers/controller.go
+++ b/pkg/workflow/controller/controllers/controller.go
@@ -51,7 +51,7 @@ type Event struct {
 func (c *Controller) Run(threadiness int, stopCh <-chan struct{}) {
 	defer c.queue.ShutDown()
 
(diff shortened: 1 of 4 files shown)


hyy0322:
LGTM

PR #1481, opened 2020-08-23
feat: support sbt accelerate

hyy0322:

Add your description

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/biz/accelerator/accelerate.go
+++ b/pkg/server/biz/accelerator/accelerate.go
@@ -48,6 +48,8 @@ func NewAccelerator(tenant, project string, wfr *v1alpha1.WorkflowRun) *Accelera
 // - '/root/.m2'  maven dependency path
 // - '/root/.gradle'  gradle dependency path
 // - '/root/.npm'  npm dependency path
+// - '/root/.ivy2' sbt default dependency cache path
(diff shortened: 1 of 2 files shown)


pkg/server/biz/accelerator/accelerate.go
```diff
@@ -48,6 +48,9 @@ func NewAccelerator(tenant, project string, wfr *v1alpha1.WorkflowRun) *Accelera
 // - '/root/.m2'  maven dependency path
 // - '/root/.gradle'  gradle dependency path
 // - '/root/.npm'  npm dependency path
+// - '/root/.sbt' sbt default configuration path
```
zhujian7, line 51: Is `the configuration path` necessary, I think the sizes of configuration files will not be very large usually.

hyy0322: I was considering configure repositories in path ```.sbt```. After some dig, I found it can be configured by sbt command line ```sbt -Dsbt.override.build.repos=true sbt -Dsbt.repository.config=<repositories-config-file-path>```.

Refer: https://stackoverflow.com/questions/3770125/how-do-i-get-sbt-to-use-a-local-maven-proxy-repository-nexus?rq=1

So, I will remove ```.sbt``` path

PR #1482, opened 2020-08-25
fix: optimize git clone process

hyy0322:

- clone with depth flag first, if fail (some version of bitbucket not support), retry without depth flag.

Fixes #

Reference to #

/cc @zhujian7 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/build/resolver/git/entrypoint.sh
+++ b/build/resolver/git/entrypoint.sh
@@ -57,12 +57,8 @@ if [ -z "${SCM_REVISION}" ]; then echo "SCM_REVISION is unset"; exit 1; fi
 if [ -z "${SCM_AUTH}" ]; then echo "WARN: SCM_AUTH is unset"; fi
 if [ "${SCM_TYPE}" = "Bitbucket" ] && [ -z "${SCM_USER}" ]; then echo "WARN: SCM_USER is required when SCM_TYPE is Bitbucket"; fi
 
-# Git clone with "--depth" option will fail when the server is Bitbucket which version less than 
-# v0.6.4(This version is not guaranteed to be accurate, I tested v0.6.4 support "--depth", but v0.5.4.9 not support)
-if [ "${SCM_TYPE}" != "Bitbucket" ]; then
-    GIT_DEPTH_OPTION="--depth=1"
-    GIT_DEPTH_OPTION_DEEPER="--depth=30"
-fi
+GIT_DEPTH_OPTION="--depth=1"
+GIT_DEPTH_OPTION_DEEPER="--depth=30"
 
 # If SCM_REPO is provided, embed it to SCM_URL
 if [ ! -z "${SCM_REPO}" ]; then
@@ -128,6 +124,8 @@ modifyURL() {
 }
 
 wrapPull() {
+    local count
+    local res=0
     # If there is already data, we should just wait for the data to be ready.
     if [ -e $WORKDIR/data ]; then
         echo "Found data, wait it to be ready..."
@@ -144,7 +142,13 @@ wrapPull() {
         failed=$(mkdir $PULLING_LOCK > /dev/null 2>&1 || echo fail)
         if [[ $failed != "fail" ]]; then
             echo "Got the lock, start to pulling..."
-            pull
+            pull || res=$?
+            if [ $res -eq 99 ] && [ "${SCM_TYPE}" == "Bitbucket" ]; then
+                echo "Fail to pull with depth flag, retry without depth flag..."
+                unset GIT_DEPTH_OPTION
+                unset GIT_DEPTH_OPTION_DEEPER
+                pull
+            fi
         else
             echo "Failed to get the lock, wait others to finish pulling..."
             while [ -d $PULLING_LOCK ]
@@ -157,7 +161,7 @@ wrapPull() {
     # Write commit id to output file, which will be collected by Cyclone
     cd $WORKDIR/data
     echo "Collect commit id to result file /__result__ ..."
-    echo "LastCommitID:`git log -n 1 --pretty=format:"%H"`" >> /__result__;
+    echo "LastCommitID:`git log -n 1 --pretty=format:"%H"`" > /__result__;
     cat /__result__;
 }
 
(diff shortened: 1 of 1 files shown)


build/resolver/git/entrypoint.sh
```diff
@@ -144,7 +141,22 @@ wrapPull() {
         failed=$(mkdir $PULLING_LOCK > /dev/null 2>&1 || echo fail)
         if [[ $failed != "fail" ]]; then
             echo "Got the lock, start to pulling..."
-            pull
+            if ! pull; then
```
zhujian7, line 144: This error judgment should be fine-tuned, only *bitbucket* server git clone with a *certain* depth-caused error will be retried.

hyy0322:
Refactor as what we talked about earlier, PTAL again. @zhujian7

build/resolver/git/entrypoint.sh
```diff
@@ -144,7 +142,22 @@ wrapPull() {
         failed=$(mkdir $PULLING_LOCK > /dev/null 2>&1 || echo fail)
         if [[ $failed != "fail" ]]; then
             echo "Got the lock, start to pulling..."
-            pull
+            pull || res=$?
+            if [ $res -eq 99 ]; then
+                echo "Fail to pull with depth flag, retry without depth flag..."
+                unset GIT_DEPTH_OPTION
+                unset GIT_DEPTH_OPTION_DEEPER
+                count=3
```
zhujian7, line 150: I think retry once is enough.

build/resolver/git/entrypoint.sh
```diff
@@ -200,18 +213,30 @@ pull() {
...
         # give it a 'oauth2:' prefix to get 'oauth2:<encoded_token>'.
         if [ ! -z ${SCM_AUTH+x} ]; then
-            SCM_URL=$( modifyURL $SCM_URL $SCM_AUTH )
+            SCM_URL_MODIFIED=$( modifyURL $SCM_URL $SCM_AUTH )
         fi
 
         if [[ "${SOURCE_BRANCH}" == "${TARGET_BRANCH}" ]]; then
             echo "Clone $SOURCE_BRANCH..."
-            git clone -v -b master ${GIT_DEPTH_OPTION:-} --single-branch --recursive ${SCM_URL} data
+            git clone -v -b master ${GIT_DEPTH_OPTION:-} --single-branch --recursive ${SCM_URL_MODIFIED} data 2>&1 | tee clone_result.log
```
zhujian7, line 221: Please delete the "clone_result.log" file before exit

build/resolver/git/entrypoint.sh
```diff
@@ -200,18 +213,30 @@ pull() {
...
         if [ ! -z ${SCM_AUTH+x} ]; then
-            SCM_URL=$( modifyURL $SCM_URL $SCM_AUTH )
+            SCM_URL_MODIFIED=$( modifyURL $SCM_URL $SCM_AUTH )
         fi
 
         if [[ "${SOURCE_BRANCH}" == "${TARGET_BRANCH}" ]]; then
             echo "Clone $SOURCE_BRANCH..."
-            git clone -v -b master ${GIT_DEPTH_OPTION:-} --single-branch --recursive ${SCM_URL} data
+            git clone -v -b master ${GIT_DEPTH_OPTION:-} --single-branch --recursive ${SCM_URL_MODIFIED} data 2>&1 | tee clone_result.log 
+            if grep "The requested URL returned error: 500" clone_result.log > /dev/null 2>&1; then
```
zhujian7, line 222: Make this code block `if ..... fi` to a common func?

build/resolver/git/entrypoint.sh
```diff
@@ -200,18 +213,30 @@ pull() {
...
+                # use code 99 for git clone --depth fail
+                return 99
```
zhujian7, line 224: The `99` `98` return codes are unreadable, any other `best practice` we can refer to?

build/resolver/git/entrypoint.sh
hyy0322: done

build/resolver/git/entrypoint.sh
hyy0322: done

build/resolver/git/entrypoint.sh
hyy0322: Shell scripts use int as return code. I think it is as I have already detailed code meaning in comments.

build/resolver/git/entrypoint.sh
```diff
@@ -157,7 +170,7 @@ wrapPull() {
     # Write commit id to output file, which will be collected by Cyclone
     cd $WORKDIR/data
     echo "Collect commit id to result file /__result__ ..."
-    echo "LastCommitID:`git log -n 1 --pretty=format:"%H"`" >> /__result__;
+    echo "LastCommitID:`git log -n 1 --pretty=format:"%H"`" > /__result__;
```
zhujian7, line 173: Why do we need to change this?

zhujian7:
And can we make this change only take effect for BitBucket, maybe some code like:
```
if [ "${SCM_TYPE}" != "Bitbucket" ]; then
```

build/resolver/git/entrypoint.sh
```diff
@@ -144,7 +142,22 @@ wrapPull() {
         failed=$(mkdir $PULLING_LOCK > /dev/null 2>&1 || echo fail)
         if [[ $failed != "fail" ]]; then
             echo "Got the lock, start to pulling..."
-            pull
+            pull || res=$?
+            if [ $res -eq 99 ]; then
+                echo "Fail to pull with depth flag, retry without depth flag..."
+                unset GIT_DEPTH_OPTION
+                unset GIT_DEPTH_OPTION_DEEPER
+                count=2
```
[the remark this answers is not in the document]
hyy0322: conut=2 means pull with GIT_DEPTH_OPTION_DEEPER twice. It means pull with GIT_DEPTH_OPTION_DEEPER retry one time.

build/resolver/git/entrypoint.sh
hyy0322: Don't need to append commit info to __result__. Just write to it is enough.

hyy0322:
> And can we make this change only take effect for BitBucket, maybe some code like:
> 
> ```
> if [ "${SCM_TYPE}" != "Bitbucket" ]; then
> ```

I think it's ok. Retry logic can fit all SCM. Though it only take effect on bitbucket. If we code like
```if [ "${SCM_TYPE}" != "Bitbucket" ]; then```,
there may be redundant code like
```
if [ "${SCM_TYPE}" != "Bitbucket" ]; then
  pull
  retry
else
  pull
fi
```

PR #1485, opened 2020-09-03
feat: support bitbucket triggers workflows by hooks

zhujian7:
BitBucket version smaller than 5.4 can not support webhooks, we
suppose these bitbucket can use hooks, like ScriptRunner, to send
a post request to cyclone server when some events happened, and
the cyclone server will handle the request and trigger workflows.

Add your description

Fixes #

Reference to #

/cc @hyy0322 @supereagle 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler/v1alpha1/webhook.go
+++ b/pkg/server/handler/v1alpha1/webhook.go
@@ -21,6 +21,7 @@ import (
 	"github.com/caicloud/cyclone/pkg/server/biz/scm/gitlab"
 	"github.com/caicloud/cyclone/pkg/server/biz/scm/svn"
 	"github.com/caicloud/cyclone/pkg/server/common"
+	"github.com/caicloud/cyclone/pkg/server/config"
 	"github.com/caicloud/cyclone/pkg/server/handler"
 	"github.com/caicloud/cyclone/pkg/util/cerr"
 )
@@ -45,6 +46,11 @@ func HandleWebhook(ctx context.Context, tenant, eventType, integration string) (
 	}
 	request := service.HTTPContextFrom(ctx).Request()
 
+	type EventDataWithIntegration struct {
+		*scm.EventData
+		Integration string
+	}
+	var allData = make(map[string][]*EventDataWithIntegration)
 	var data *scm.EventData
 
 	if request.Header.Get(github.EventTypeHeader) != "" {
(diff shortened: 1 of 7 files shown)


pkg/server/handler/v1alpha1/webhook.go
```diff
@@ -60,34 +66,91 @@ func HandleWebhook(ctx context.Context, tenant, eventType, integration string) (
...
-		if err != nil {
-			return newWebhookResponse(err.Error()), err
+		hookEvent := request.Header.Get(bitbucket.HookEventHeader)
+		tenants := make([]string, 0)
+		var terr error
+		if hookEvent == "true" {
+			serverAddress := strings.Trim(request.Header.Get(bitbucket.ServerAddressHeader), "/")
+			hookData := bitbucket.ParseHookEvent(request)
+
+			// Trigger workflows for all tenants
```
hyy0322, line 76: Why?

pkg/server/handler/v1alpha1/webhook.go
```diff
@@ -45,6 +46,11 @@ func HandleWebhook(ctx context.Context, tenant, eventType, integration string) (
...
 	request := service.HTTPContextFrom(ctx).Request()
 
+	type EventDataWithIntegration struct {
+		*scm.EventData
+		Integration string
+	}
+	var allData = make(map[string][]*EventDataWithIntegration)
 	var data *scm.EventData
 
 	if request.Header.Get(github.EventTypeHeader) != "" {
```
hyy0322, line 56: Can we use switch case for this four types SCM

pkg/server/handler/v1alpha1/webhook.go
hyy0322, line 76: Why?

pkg/server/handler/v1alpha1/webhook.go
hyy0322, line 56: Can we use switch case for this four types SCM ?

zhujian7: No, the keys in the request header are different for every SCM.
- github.EventTypeHeader
- bitbucket.EventTypeHeader
- svn.EventTypeHeader

pkg/server/handler/v1alpha1/webhook.go
zhujian7: The hook configured in the bitbucket server can not know which tenant will be triggered.

zhujian7:
@hyy0322 PTAL

PR #1500, opened 2020-09-16
fix: validate scm auth info problem

zhujian7:
only check token type auth, also need to check password type auth.

Add your description

Fixes #

Reference to #

/cc @hyy0322 @supereagle 

**Release note**:

```release-note
NONE
```


pkg/server/biz/scm/scm.go
```diff
@@ -125,16 +125,21 @@ func GenerateSCMToken(config *v1alpha1.SCMSource) error {
 
 	if generatedToken != "" && config.AuthType == v1alpha1.AuthTypePassword {
```
hyy0322, line 126: May be ```generatedToken != ""``` is enough?

PR #1517, opened 2020-11-02
fix(bitbucket): correct webhook event

knight42:

We actually want to trigger pipeline on `pr:from_ref_updated` event
rather than `pr:modified`.

Fixes #

Reference to #

/cc @zhujian7 

**Release note**:

```release-note
bitbucket: fix a bug that PR commits won't trigger pipeline.
```

### DIFF
--- a/pkg/server/biz/scm/bitbucket/server/repos_webhooks.go
+++ b/pkg/server/biz/scm/bitbucket/server/repos_webhooks.go
@@ -12,9 +12,11 @@ const (
 	RefsChanged = "repo:refs_changed"
 	// PrOpened is the event key about pull request opened.
 	PrOpened = "pr:opened"
-	// PrModified is the event key about pull request modified.
+	// PrModified means a pull request's description, title, or target branch is changed.
 	// It be supported after the version is 5.10. Ref to: https://confluence.atlassian.com/bitbucketserver/bitbucket-server-5-10-release-notes-948214779.html
 	PrModified = "pr:modified"
+	// PrFromRefUpdated means a pull request's source branch has been updated.
+	PrFromRefUpdated = "pr:from_ref_updated"
 	// PrCommentAdded is the event key about a comment added on the pull request.
 	PrCommentAdded = "pr:comment:added"
 )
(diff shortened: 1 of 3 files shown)


pkg/server/biz/scm/bitbucket/server/server.go
```diff
@@ -334,19 +334,7 @@ func (b *BitbucketServer) CreateWebhook(repo string, webhook *scm.Webhook) error
 		for _, e := range webhook.Events {
 			switch e {
 			case scm.PullRequestEventType:
-				version, err := GetBitbucketVersion(b.v1Client.client, b.v1Client.baseURL)
-				if err != nil {
-					return err
-				}
-				isSupportPrModified, err := IsHigherVersion(version, SupportPrModifiedEvent)
```
zhujian7, line 341: Why delete this?

knight42: because we don't need to check the version now.

knight42: I think we only want to check the version if we want to know if `PrModifiedEvent` is supported, but now `PrModifiedEvent` is not used anymore, so I delete these lines.

pkg/server/biz/scm/bitbucket/server/repos_webhooks.go
```diff
@@ -12,9 +12,11 @@ const (
 	RefsChanged = "repo:refs_changed"
 	// PrOpened is the event key about pull request opened.
 	PrOpened = "pr:opened"
-	// PrModified is the event key about pull request modified.
+	// PrModified means a pull request's description, title, or target branch is changed.
```
zhujian7, line 15: If the target branch changed, do we need to trigger the pipeline?

knight42: I think we need to trigger the pipeline, as the code on the new target branch might be different with the old one.

zhujian7: I agree, so do we need to create the webhook with both types `PrModified` and `PrFromRefUpdated`?

knight42: but could we tell the modification of target branch from PR's description or title? 

IMHO PR's target branch rarely changes, while its title or description changes much more often. I guess we don't want to trigger the pipeline every time the title or description changes.

zhujian7: I think we could keep the behavior the same as before.

PR #1518, opened 2020-11-03
feat(workflow): set the weight of cpu request of custom containers

knight42:

* Add a field `WorkloadCustomContainerCPUWeight` to `pkg/server/config.CycloneServerConfig`
* `applyResourceRequirements` would adjust the cpu request of the containers according to the above setting

Fixes #

Reference to #

/cc @zhujian7 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/workload/pod/builder.go
+++ b/pkg/workflow/workload/pod/builder.go
@@ -1013,7 +1013,11 @@ func applyResourceRequirements(containers []corev1.Container, requirements *core
 				if common.OnlyCustomContainer(c.Name) {
 					c.Resources.Limits[k] = v
 				} else {
-					c.Resources.Limits[k] = zeroQuantity
+					if k == corev1.ResourceCPU {
+						c.Resources.Limits[k] = v
+					} else {
+						c.Resources.Limits[k] = zeroQuantity
+					}
 				}
 			}
 		}
--- a/pkg/workflow/workload/pod/builder.go
+++ b/pkg/workflow/workload/pod/builder.go
@@ -929,21 +929,54 @@ func divideQuantity(quantity resource.Quantity, n int) (resource.Quantity, error
 //   consume negligible resources.
 // - the other containers(workload containers or custom containers, there is only one at most of the time)
 //   will average the pod resource requirements
-func applyResourceRequirements(containers []corev1.Container, requirements *corev1.ResourceRequirements, averageToContainers bool) []corev1.Container {
-	var results []corev1.Container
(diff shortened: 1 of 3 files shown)


pkg/server/config/config.go
```diff
@@ -75,6 +75,8 @@ type CycloneServerConfig struct {
 
 	// Artifact config for artifacts which are managed by cyclone server
 	Artifact ArtifactConfig `json:"artifact"`
+
+	WorkloadCustomContainerCPUWeight int `json:"workload_custom_container_cpu_weight"`
```
zhujian7, line 79: Please add some comments for the field

pkg/workflow/workload/pod/builder.go
```diff
@@ -994,8 +1036,8 @@ func (m *Builder) ApplyResourceRequirements() error {
 		requirements = m.wf.Spec.Resources
 	}
 
-	m.pod.Spec.InitContainers = applyResourceRequirements(m.pod.Spec.InitContainers, requirements, false)
-	m.pod.Spec.Containers = applyResourceRequirements(m.pod.Spec.Containers, requirements, true)
+	m.pod.Spec.InitContainers = applyResourceRequirements(m.pod.Spec.InitContainers, requirements, false, config.Config.WorkloadCustomContainerCPUWeight)
```
zhujian7, line 1039: Init containers executes one by one, so I think it is not necessary to comply with this rule.

zhujian7 (changes requested):

knight42:
@zhujian7 PTAL

pkg/server/config/config.go
```diff
@@ -75,6 +75,10 @@ type CycloneServerConfig struct {
 
 	// Artifact config for artifacts which are managed by cyclone server
 	Artifact ArtifactConfig `json:"artifact"`
+
+	// WorkloadCustomContainerCPUWeight represents the proportion of cpu resource that custom containers would consume(1-100).
+	// If this field is equal to 0, its value is changed to 100.
+	WorkloadCustomContainerCPUWeight int `json:"workload_custom_container_cpu_weight"`
```
zhujian7, line 81: `CustomContainerCPUWeight` is simpler?

zhujian7 (approved):

zhujian7:
@knight42 please fix the CI.

knight42:
/hold
Leave time for review. @zhujian7 PTAL

knight42:
/close

b/c this is not the root cause.

PR #1523, opened 2020-11-16
Make coordinator independent of docker

knight42:

* BREAKING CHANGE: change the path of `__result__` file to `/cyclone/results/__result__`
* feature: make coordinator independent of docker 

Fixes #

Reference to #

/cc @zhujian7 

**Release note**:

```release-note
The path of `__result__` file has been changed to `/cyclone/results/__result__`
```


pkg/workflow/workload/pod/builder.go
```diff
@@ -452,6 +461,11 @@ func (m *Builder) ResolveInputResources() error {
 					Name:      common.ToolsVolume,
 					MountPath: common.ToolboxPath,
 				},
+				{
+					Name:      common.CoordinatorSidecarVolumeName,
```
zhujian7, line 465: Why does input container need this volume?

pkg/workflow/workload/pod/builder.go
```diff
@@ -822,8 +847,9 @@ func (m *Builder) AddCoordinator() error {
 				MountPath: common.CoordinatorResolverPath,
 			},
 			{
-				Name:      common.HostDockerSockVolumeName,
-				MountPath: common.DockerSockFilePath,
+				Name:      common.CoordinatorSidecarVolumeName,
+				MountPath: filepath.Join(common.CoordinatorWorkspacePath, "results"),
+				SubPath:   "__results__",
```
zhujian7, line 852: Hold the `__results__` as a common const var?

PR #1525, opened 2020-11-19
fix(workflow): remove unnecessary volume from coordinator

knight42:

coordinator no longer depends on Docker now.

Fixes #

Reference to #

/cc @zhujian7 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/workflow/workload/pod/builder.go
+++ b/pkg/workflow/workload/pod/builder.go
@@ -236,16 +236,6 @@ func (m *Builder) CreateVolumes() error {
 		},
 	})
 
-	// Add host volume for host docker socket file for coordinator container.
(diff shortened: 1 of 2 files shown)


pkg/workflow/workload/pod/builder.go
```diff
@@ -236,16 +236,6 @@ func (m *Builder) CreateVolumes() error {
 		},
 	})
 
-	// Add host volume for host docker socket file for coordinator container.
-	m.pod.Spec.Volumes = append(m.pod.Spec.Volumes, corev1.Volume{
-		Name: common.HostDockerSockVolumeName,
```
zhujian7, line 241: Can we delete `common.HostDockerSockVolumeName` or docker related const now?

knight42: removed

PR #1529, opened 2020-11-30
fix: cluster lost should not block close cluster

zhujian7:
**What type of PR is this?**

cluster lost should not block close cluster

Fixes #

Reference to #

/cc @knight42 

**Does this PR introduce a user-facing change?**:

```release-note
Cluster lost does not block close cluster
```

### DIFF
--- a/pkg/util/cerr/error.go
+++ b/pkg/util/cerr/error.go
@@ -184,25 +184,13 @@ func AutoAnalyse(err error) error {
 
 // TCPConnectionError check whether the err belongs to tcp connection problem.
 func TCPConnectionError(err error) bool {
-	if err == nil {
+	e := err.Error()
+	if !strings.Contains(e, "dial tcp") {
 		return false
 	}
 
-	if strings.Contains(err.Error(), "dial tcp") && strings.Contains(err.Error(), "connect: connection refused") {
-		return true
-	}
-
-	if strings.Contains(err.Error(), "dial tcp") && strings.Contains(err.Error(), "no such host") {
-		return true
-	}
-
-	if strings.Contains(err.Error(), "dial tcp") && strings.Contains(err.Error(), "i/o timeout") {
-		return true
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/util/cerr/error.go
```diff
@@ -181,3 +181,28 @@ func AutoAnalyse(err error) error {
 
 	return err
 }
+
+// TCPConnectionError check whether the err belongs to tcp connection problem.
+func TCPConnectionError(err error) bool {
+	if err == nil {
```
knight42, line 187: Is it necessary to check if `err` is nil here?

pkg/util/cerr/error.go
```diff
@@ -181,3 +181,28 @@ func AutoAnalyse(err error) error {
...
+		return false
+	}
+
+	if strings.Contains(err.Error(), "dial tcp") && strings.Contains(err.Error(), "connect: connection refused") {
```
knight42, line 191: looks like we could save `err.Error()` to a variable.

knight42 (approved):

zhujian7:
@knight42 PTAL

pkg/util/cerr/error.go
```diff
@@ -181,3 +181,20 @@ func AutoAnalyse(err error) error {
...
 }
+
+// TCPConnectionError check whether the err belongs to tcp connection problem.
+func TCPConnectionError(err error) bool {
+	e := err.Error()
+	if !strings.Contains(e, "dial tcp"){
+		return false
+	}
+
+	if strings.Contains(e, "connect: connection refused") ||
```
knight42, line 192: nit:
```suggestion
	return strings.Contains(e, "connect: connection refused") ||
		strings.Contains(e, "no such host")||
		strings.Contains(e, "i/o timeout")||
		strings.Contains(e, "connect: connection timed out")
```

knight42 (approved):

pkg/util/cerr/error.go
zhujian7: Fixed

PR #1560, opened 2021-01-20
fix: only set LabelWorkflowRunPRRef when needed and correct storageClass

knight42:
**What type of PR is this?**

/area cicd
/kind cleanup

* Only set LabelWorkflowRunPRRef when needed(e.g. we don't need to set this label if the workflorRun is trigger by tag release)
* Correct storageClass name

Fixes #

Reference to #

/cc @zhujian7 

**Does this PR introduce a user-facing change?**:

```release-note
* Only set LabelWorkflowRunPRRef if workflowRun is triggered by PR
* Change storageClass name to `platform-storageclass`
```

### DIFF
--- a/manifests/cyclone/templates/server.yaml
+++ b/manifests/cyclone/templates/server.yaml
@@ -193,5 +193,5 @@ spec:
       storage: 10Gi
     requests:
       storage: 10Gi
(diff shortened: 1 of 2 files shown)


zhujian7:
cc @hyy0322

